PR #18, opened 2019-06-26
Revise Datasets section

adswa:
I'm trying to set a 'hands-on' mood already in this section. The idea is to have readers see (and experience, if they want to) what a dataset is and can do, because I found it hard to stay abstract.

The current idea is to have the [studyforrest-data-phase2](https://github.com/psychoinformatics-de/studyforrest-data-phase2) as one example dataset to display some cool dataset properties. 

Currently, I'm using screenshots to showcase everything, because the syntax highlighting is better than the ``code-block:: bash`` section, and commands are very short and easy to type, should anyone want to follow along. Not sure whether I might revert this to ``code-block::`` sections later...


mih:
There is a strong reason to not use screenshots: we cannot check if they still work, or search for them. Let's talk what the syntax highlighting isn't doing, and configure it better.

adswa:
Will do, thx!

adswa:
You're extension is amazing to work with, @mih. I'll push my changes, but there is a problem:

because I'm installing a dataset with the autorunrecord directive in the workdir, I'm setting up all of the .git and .git-annex stuff in the background. A make clean afterwards fails due to permission errors:

```sh

(handbook) ╭─adina@muninn ~/repos/datalad-handbook on ds!
╰─➤ make clean-examples clean
find docs -name _examples -type d | xargs rm -rf
cd docs && make clean
make[1]: Entering directory '/home/adina/repos/datalad-handbook/docs'
rm -rf _build/*
rm: cannot remove '_build/wdirs/intro/datasets/studyforrest-data-phase2/.git/annex/objects/Kj/jg/MD5E-s64551297--8f3a805295866b61ebf31a2f74e17d57.nii.gz/MD5E-s64551297--8f3a805295866b61ebf31a2f74e17d57.nii.gz': Permission denied
make[1]: *** [Makefile:45: clean] Error 1
make[1]: Leaving directory '/home/adina/repos/datalad-handbook/docs'
make: *** [Makefile:7: clean] Error 2

``

adswa:
mhhh ignore this for now, I have to clean up a bit

mih:
> because I'm installing a dataset with the autorunrecord directive in the workdir, I'm setting up all of the .git and .git-annex stuff in the background. A make clean afterwards fails due to permission errors

Ah right -- I should have anticipated that. Will push a fix.

adswa:
thx!

mih:
Fix pushed into this branch!

adswa:
thx, this makes it work for me! Locally, everything builds and looks as it should. Travis is unhappy, though. ``datalad install`` does not succeed, and everything onwards then fails. Do you know any way how I can look into what is going on in for example ``docs/intro/_examples/dataset  `` *in this remote*?

adswa:
yay :-)

mih:
Ah! The idea is to actually commit the locally generated files under `_examples`. If they are present in the repo, travis will not attempt to run anything and datalad is not needed. This way we can run arbitrary code locally, and only use travis to check whether the docs build properly -- and not if that code would run. Running everything would require a long time very quickly and also require a more complex setup.

So if you commit the generated file, you can revert the dependency change.

Fingers crossed...

mih:
FTR: Will make an attempt to rectify this now (working on a related improvement)...

mih:
Looks good, will merge.

adswa:
Thx!

PR #19, opened 2019-06-26
Installation instructions in the Introduction section

adswa:
Start a section on how to install DataLad.
So far the content is copy-pasted from the [docs](docs.datalad.org) and the [website](datalad.org).

Some TODOs concern
- adding some information (I'm not sure what the current status is, but at least a "it does not work, you can use a virtual machine, ...") for Windows users and 
- a less intimidating way of talking about non-admin installation of git-annex

### DIFF
--- a/docs/contents.rst.inc
+++ b/docs/contents.rst.inc
@@ -8,6 +8,7 @@ What DataLad is all about...
 
    intro/duction
    intro/philosophy
+   intro/installation
    intro/datasets
    intro/contributing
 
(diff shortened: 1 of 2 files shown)


mih:
For windows people: it does work somewhat. Core functionality should be OK (current main focus of this effort), and is covered by tests. There are numerous issues with SSH connections on windows, though. It works much better on Win10 within the linux subsystem, and this is also where the hopes for the future are focused on. DataLad generally feels sluggish on Windows, because of a range of filesystem issues that also affect Git itself.

PR #24, opened 2019-06-27
[WIP] Test of using a sphinx extension for auto-recorded examples

mih:
This adds a new chapter "remove" to demonstrate the use of a custom sphinx extension to run arbitrary (shell) code and capture the output. The goal is to be able to auto-execute real examples, but capture and cache their output for inclusion in the book, without having to run all code at build-time.

Here is the extension:

https://github.com/mih/autorunrecord

It should be pip-installable: `pip install autorunrecord`

The rest should be automatic.

The `runrecord` directive shown in the examples can be used in the same fashion and has the same features as the `literalinclude` directive that comes with sphinx (in fact, it is based on it). In addition, however, it can auto-generate the to-be-included files based on the code given in the block. Moreover, if it is necessary (e.g. for didactic reasons) to simplify the displayed command, but a more complicated variant is needed for optimal output, such a command can be given via the `realcommand` option.

All examples in a given RST file run sequentially in a dedicated work dir, hence incremental examples are possible.

Additional features can be and will be added later, such as:

- `prep` commands that establish a particular state of a worktree, but have no rendered equivalent
- `pretest`/`posttest` option that allow for specification of arbitrary test code that can verify preconditions and expected command results -- to by able to use these examples as doctests

TODO:

- [x] figure out what is missing for travis to like this

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ sudo: false
 language: python
 python:
   - "3.6"
-install: pip install sphinx
+install: pip install -r requirements.txt
 script:
   - make doctest
   - make html
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,11 @@
-.PHONY: build
+.PHONY: build clean-examples
 build: html
 
 # this pattern rule lets you run "make build" (or any other target
 # in docs/Makefile) in this directory as though you were in docs/
 %:
 	cd docs && make $@
+
+# wipe out all recorded examples
+clean-examples:
+	find docs -name _examples -type d | xargs rm -rf
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -32,6 +32,7 @@ extensions = [
     'sphinx.ext.todo',
     'sphinx.ext.intersphinx',
     'sphinx.ext.doctest',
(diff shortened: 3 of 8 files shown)


mih:
@loj @adswa This should be working now. Please give it a try locally. This would be needed to run (all in the root of the repo):

```
# install all needed dependency into your local environment
pip install -r requirements.txt

# wipe out the recorded examples that I commited to this PR
make clean-examples clean

# rebuild from scratch
make
```

This should give you back one identical file, and a date difference for the second (check with a plain `git diff`). If that is the case, it is working as intended and the PR could be merged.

Once in, we have to see whether the current functionality is what is needed, or if we need to make changes beyond the issues outlined in the top comment.

mih:
FTR: The work directory for all examples from a single RST file can be found under `docs/_build/wdirs` plus the relative path of the RST file. A `make clean` will wipe them out, but there is no need to keep them around, as the captured outputs will prevent any code execution, as long as the respective text file exists.

adswa:
Thanks a lot for this! It works as intended, so I'll merge it, and play around with it.

PR #37, opened 2019-07-01
Opinionated heading standardization -> book structure

mih:
This move uniformly adopts the conventions summarized in the sphinx docs (https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html?highlight=chapter#sections). While it really is arbitrary, I think there is value in doing it uniformly.

I also introduce a uniform .../index.rst file for all chapters that can host the preamble.

There is still an open issue in that sphinx treats files as more than just arbitrary structures, i.e. whatever headline definition is used, the top-most one in each file becomes the title (etc.). This is also reflected in the PDF -- each file represents a "chapter", hence conflicts with the TOC defined in contensts.rst.inc.

We should standardize this as well, to make it less confusing.

One way would be to really consider each file to be a chapter and treat the current "chapters" (as defined in the TOC) as "parts" of the book -- which also makes sense to me. If that shall be the way, I'd propose to frame the introduction a bit differently (to make the difference to "basics" more pronounced -- pretty much as a guide to the book itself. It could have "why should I read this?" or "for whom which part this is written", "what do I need to work with the book", "how can I contribute". Bits like "philosophy" would then possibly need to be moved into the "basics" part.

Depending on the feedback to this I would rework this change.


adswa:
I like the structure of "parts" being ``Introduction``, ``Basics``, ``Use cases``, with the files "underneath" them being either ``chapters`` (in Sphinx terminology) or ``sections``. TOCs and (sub)headers as you changed them in here create an easily comprehensible structure. Standardizing how we define headers is a very good idea. Thanks for the pointer to an explanation of the different header types and sphinx interpretation of them, up to now, I did those rather randomly.

ATM I think of the `` The Introduction`` as being the part to tell everyone what DataLad is (I think we can't take any specific knowledge about DataLad as granted at all), a way to "advertise" what is great about it in general to motivate further reading, and to incorporate "how to read and work with this + contribute" information -- so I agree to the content you pointed out:

> It could have "why should I read this?" or "for whom which part this is written", "what do I need to work with the book", "how can I contribute".

(However, I like the ``contributing.rst`` in the appendix, so a general motivation to contribute and a link to the appendix would suffice IMO). 
I am unsure about moving the Philosophy into ``Basics``. Personally, I envision the Philosophy to be more of a general overview/mindset that introduces DataLad from an "ideational" instead of procedural point of view. ``The Basics`` one the other hand, in my opinion, can be procedural, non-abstract workflows, commands, or concepts. However, this is a vague concept/opinion in my head, and I think that for the upcoming weeks, we, the book, and its content are flexible enough to just try things out. I'd be happy to restructure in the way you propose, and for parts where it might not appear optimal afterwards, we can adjust.

mih:
@adswa Thx. I think your scenario is a compromise between the two extremes, and it sounds good to me. Let's wait for @loj to have an odd number of opinions ;-)

loj:
> [adswa wrote]
> I like the structure of "chapters" being Introduction, Basics, Use cases, Appendix, with the files 
> "underneath" them being sections. 

I also like this and think it makes a lot of sense.

I agree with making `Introduction` a sort of guide for how to approach the book. This makes me lean away from including Philosophy as part of it. However, I see Adina's point that the Philosophy section is an overview/mindset approach rather than procedural, so it feels a bit odd putting it in `Basics`. Right now I lean towards moving Philosophy to `Basics`, so that we can keep the focus of `Introduction` on approach to the book. Then adjust as necessary as the book evolves.

> [adswa wrote]
> (However, I like the contributing.rst in the appendix, so a general motivation to contribute and a link to 
> the appendix would suffice IMO).

Good idea!

mih:
Thinking...

That makes two options. The intro is:

1. an intro to the book
2. an intro to datalad

If (1) and if 'basics' evolves to be (a collection of) introductory pieces on various related technologies (shell, git, git-annex, ...) it could be a fairly sizable chunk of content before any word is spoken about Datalad -- that is likely not a good thing. So I am now leaning towards (2).

If that is the path, and thinking further, 'installation' is something 'basic', and could move there. It would make sense to me, because it is consistent with a difference between basics and introduction being the aspect of procedural vs conceptual information. Installation instructions are clearly of the former type.

'Contributing' is essentially metainformation and can live anywhere. We could even have it be in plain Markdown and link to its location on Github (i.e. not have it be part of the book at all). Contributions would need to come this way anyways, and this way these docs are tightly bound to the channel they describe.

I don't have my opinion fully formed yet...

adswa:
I agree that installation instructions can live in Basics. 

I also think that any non-DataLad related intro (whats currently our ``chapter 0``) should be as concise as possible, and should not talk about ``git`` or ``git-annex`` at all. But I believe its good to have a resource on anything Unix necessary to understand the example code close by, i.e. *in* the book, not as references to outside resources. If we want to, we could put this information (``chapter 0``) in the end of ``Basics`` or the ``Appendix`` though, with a reference to read it first, should one be not used to working in a shell. This way, no one will stumble across it accidentally and then get bored. I believe for this some of the institutes PhD students would be a great first audience (I'm in contact with Marisa and Lisa, and they are happy to provide feedback) to give us an idea of which and how many non-DataLad commands are helpful.

Instead of a ``how to contribute`` section a ``CONTRIBUTING.md`` or longer ``README.md`` are also not a bad idea. Essentially, I have no opinion on this, its fine for me either way.

mih:
Still thinking, because the concept is not yet unambiguously clear to me  (and we are also seeing the limitations of github discussions for nailing down concepts, as opposed to bugs in code)... Here are a bunch of statements with boxes to check for approval, in case there is consensus for any of them. Feel free to edit this list and add more alternatives.

- [ ] the `introduction` is primarily an introduction to DataLad, not just an "how to read this book". But of course this is relevant information that needs to be present at the start of the book -- just not the main focus.
- [ ] the `introduction` is practically free of hands-on content, i.e. no installation instructions, no demos. Instead, it is about concepts, analogies, general problems
- [ ] in order to avoid too much of a mental split between a reader's desire to learn how to actually do things vs. conceptual information, the `introduction` is purposefully kept short and serves as a "narrated" table of contents with plenty of references to other parts of the book.
- [ ] the order of content in the `introduction` is determined by the hierarchical relationship of the explained concepts
- [ ] the `basics` chapter delivers 'hands-on'-style content on skills that are *crucial* for using DataLad productively. Any non-essential information is not in basics, but collected in an `appendix` (or elsewhere). The latter includes dedicated/standalone Git/annex tutorial/reference like content.
- [ ] installation instructions are crucial practical knowledge for DataLad use, hence they go into `basics`
- [ ] information on how to contribute to the book is non-crucial practical information, hence lives in an appendix, or elsewhere
- [ ] the order of content in `basics` is determined by the order in which they become relevant for a novice DataLad user
- [ ] `introduction` and `basics` feature (to the maximum degree possible) a common narrative and and aligned set of examples
- [ ] any other specific scenario or application involving DataLad goes into a `use cases` part (which may get a better name in the future)
- [ ] there is no inherent order in `use cases`, but chapters may be grouped by domain, skill-level, or DataLad functionality involved (or combinations of those). But see next...
- [ ] any content in `use cases` can deviate from the example/narrative used for `introduction` and `basics` whenever necessary (e.g. concrete domain specific use cases). However, if possible, common example datasets, names, terms should be adopted, and the broadest feasible target audience should be assumed. Such generic content should appear early in the `use cases`
- [ ] any chapter in `use cases` is written as a more-or-less self-contained document that makes references to `introduction` and `basics`, but only few, and more general ones to other use cases. This should help with long-term maintenance of the content, as the specifics of how to approach a particular use case optimally may change/evolve over time, and cross-references to specific functionality might become invalid. The standalone-nature of these chapters might also encourage 3rd-party contributions, as there is no need to fully integrate with the rest of the book.

adswa:
I have added boxes for each of us, so that we can establish more easily whether consensus exists.

mih:
> Adina is not sure what the "hierarchical relationship" is

Sorry, I think I went over the top and just stated something obvious: nothing else than normal introduction of terms and concepts, e.g. first say what a version control system is, and only then explain that DataLad is one. Don't even know how it could be any other way -- will edited that out now -- thanks for calling it out.

adswa:
I think there is a lot of consensus, @mih

mih:
Cool, thx for the feedback. I will implement this, and also take the material from here and populate a contributing.md file with it.

mih:
@loj @adswa From my POV this is complete now.

adswa:
Looks good to me, I will merge.

mih:
Thx!

PR #39, opened 2019-07-01
Style/Theme tweaks

mih:
I cannot claim that this is better than what was there before, but it is different ;-)

And it removes a bunch of customizations that we do not need (at least at the moment) -- and less code is objectively better.

With more content in the book we do want to think about additional customizations. For example, with a dedicated directive for git-user-note, like

```rst
.. gitusernote::

   Something sophisticated...
```

we would have better control over when to include the content where, and how it (not just look, but) would structured. For example, we could have a version rendered that does not show these notes at all.


mih:
Merge?

adswa:
Merge!

PR #44, opened 2019-07-03
Sketch of a fully custom `gitusernote` directive

mih:
This is what is needed to be able to exercise pretty much anything that we could think of in terms of presentation and behavior of these notes. At the moment it merely mimics what was there before. What is different is that all formatting is now done fully manual, so the new code exposes all relevant interfaces.

In order to achieve this, we now have a Python package for the book included in the repo.

Now we can start thinking about how to present this content best: folded, optionally fully removed from a PDF by, configuration...

In the same spirit we can also consider adding additional custom directives for the book (e.g. for exercises, ...)

WIP:
- [x] readthedocs needs to be adjusted and verified to be OK with this

When this is merged, readthedocs must be reconfigured to "install project" (under Advanced Settings), otherwise builds will start failing.

### Changes to local build setup

To be able to build locally, enter your virtualenv and the root of this repository and run `pip install -e .`


adswa:
Cool! Locally, that all works for me.

PR #47, opened 2019-07-03
PlantUML support for crazy-flexible plain-text diagrams

mih:
On Debian: `sudo apt-get install plantuml`

Redo: `pip install -r requirements.txt`

Demo page is included, many many more examples at http://plantuml.com

Rendered demo page: http://handbook.datalad.org/en/helpers/remove/uml.html (not all types are supported by the version on readthedocs).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
+dist: xenial
 language: python
 python:
   - "3.6"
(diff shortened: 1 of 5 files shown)


adswa:
FYI: installation and demos (not the last two ones) works flawlessly. These diagrams are extremely cool!

adswa:
Is this good to merge?

mih:
I think it is functional enough to be merged. If we need additional flexibility, we can attempt to make this happen later.

Issue #51, opened 2019-07-04
Ideas on how to balance file length and recreation of datasets in individual workdirs


adswa:
When recording code and its output, every single .rst file gets an individual workdir associated with it in ``docs/_build/wdirs/``.

The dataset I'm continuously building up and extending will need to be in every one of these workdirs (as many as ``.rst`` files I am writing) in order for the code to work. I don't think it is feasible to create potentially dozens of datasets, slightly increasing in size between different ``.rst`` files. But if I write down everything in a single file, I'm creating - at least for the web version of the book - a really really really long page, and in principle, I'd like to chunk it up into individual pages, and also have toc entries for each of these pages.

Does anyone have an idea on how to balance these conflicting demands? Is there a way to "pagebreak" single ``.rst`` files and create toc entries referencing specific sections in files? Or, alternatively, specify a common workdir manually?

mih:
I was already thinking of having a dedicated config option for the workdir. The examples tend to start with a `cd` already.

mih:
There is an issue, however, if things are not constrained to a single RST file, we have to deal with dependency resolution and determine an execution order. Sphinx will only touch RST files that have been edited on a rebuild, but if such a partial rebuild now requires the output of examples that will not be re-rendered, the autorunrecord extension will never get to see those steps.

In any case, there is no sense in having the structure of the book be determined by the limitations of the supporting code.

It would be good to see a concrete example if a problematic case. Maybe there is a simple solution possible. If not, we could go to some kind of two-pass update strategy, where we first collect the code bits, and then run it to update the record files. But then we would always have to build twice. Not extremely bad, but also not nice.

adswa:
Let me commit and push what I have created today, so that you could take a look what the problematic situation currently is. (give me 15 mins)

mih:
We have `workdir` for a while now. Closing

PR #52, opened 2019-07-04
Starting the Basics

adswa:
Currently, this is an illustration of a problem described in #51.
There are three new files (naming will have changed once this is not WIP anymore, and there are many more TODOs):

- ``c1.rst``: Basics on ``datalad create``
- ``ds_fill.rst``: Basics on saving new content to a dataset
- ``c2.rst``: [WIP] Basics on modifying tracked content

The files are individual files instead on one large file to create separate references in the toctree and to not have a gigantic single page (that could demotivate impatient readers). Also, it chunks learning goals a bit.

The problem is, that in ``c1.rst`` I create a dataset ``DataLad-101``, in ``ds_fill.rst`` I put content into this dataset, and in ``c2.rst`` I continue putting content inside. The workdirs of ``ds_fill.rst`` and ``c2.rst`` however do not contain the ``DataLad-101`` dataset, this only exists in the work dir of ``c1.rst``. This makes Sphinx puke, because the runrecord directives puke (when I emphasize non-existing lines for example).


mih:
Thanks. I will take a look and think about how to deal with this.

adswa:
A very ugly hack is what I just did: ``make`` the _examples with all content in a single file, then separate the files again, and make the html. But this way, you should be able to see the intended results if you build the html files from this branch.

mih:
I have implemented the `workdir` configuration option, but I have now discovered a core limitation in sphinx that will prevent us for continuing on this path: https://github.com/sphinx-doc/sphinx/issues/5940

Essentially, we will not be able guarantee the order of processing of RST files.

We can likely work around this limitation, by introducing a file naming scheme that makes the alphabetical file name order reflect their intended execution order, but that won't be a general solution -- although it might just work long and good enough.

A proper solution that is more in the spirit of sphinx could be to build a custom `Builder` https://www.sphinx-doc.org/en/master/extdev/builderapi.html#sphinx.builders.Builder that is invoked separately, and actually runs and writes out the example output.

ATM I'd say we adopt a scheme with sortable file names, and see how long that will serve us well enough.The builder solution can be attempted later, there should be no required change in the actual directive setup, at least as far as I can see for now.

Thoughts?

adswa:
I agree - lets for now sort filenames.

mih:
All changes are pushed. I checked that all examples rebuild cleanly, but I did not include the resulting changes. autorunrecord v0.3 is now needed, please update.

adswa:
Thanks a lot, this all works for me!

mih:
If you rerun all the examples, commit all the new files too, and push. The tests should pass again. I can also do that -- just let me know.

docs/basics/101-3-modify.rst
```diff
@@ -0,0 +1,48 @@
...
+   :realcommand: cd DataLad-101 && datalad status
+
+   $ datalad status
+
+Save this file in its current state:
+
+.. runrecord:: _examples/DL-101-m-3
+   :language: console
+   :workdir: dl-101
+   :realcommand: cd DataLad-101 && datalad save -m "Add notes on datalad create" notes.txt
```
adswa: @mih, @loj: I'm writing further on this, and I'm undecided on which approach to take:

Currently, I'm creating a small text file and ``datalad save`` this. The repository that gets created is a git-annex v5 repository. Therefore, this files content gets locked, and in order to modify it, I need to introduce datalad unlock and the reason behind locking.

An alternative would be to save the textfile to git with the ``to-git`` flag. This would postpone locked data content, but will make an explanation of the differences between saving and not saving to git necessary.

I believe both ways are possible. Do you have a favorite between the two? (or an alternative?)

mih: There are more then two alternatives. Few more:

- a V6 repository that gets put into an adjusted unlocked branch (`git annex adjust --unlock`). This is not the same as running `unlock`, it is more radical, but might also turn out to be the necessary thing to do, in order to avoid foreseeable issues down the line. But it is also something that we have least practical experience with -- which doesn't mean that we shouldn't acquire it.

- apply a (sensible) default configuration to the dataset. This will become relevant later anyways (when going over the YODA principles). For now it would be as simple as

  ```
  datalad create -c text2git targetdirectory
  ```
  at the very beginning. This leads to:

  ```
  % cat targetdirectory/.gitattributes

  * annex.backend=MD5E
  **/.git* annex.largefiles=nothing
  * annex.largefiles=(not(mimetype=text/*))%                                                                                                                                                   
  ```

  and ensures that any text file is tracked directly with Git (not annexed).

adswa: Thanks! A sensible default configuration currently sounds like a good choice at this moment to me, thx for bringing this option up!

adswa:
If tests pass, from my side this is ready for content-related feedback.

docs/basics/101-2-populate.rst
```diff
@@ -0,0 +1,213 @@
...
+
+    - it should be obvious: do not say nasty things about other people
+
+.. gitusernote::
+
+   Just as in Git, new files are not tracked from their creation on, but only when
+   explicitly given to Git (in Git terms with an initial ``git add``). But different
+   from the common Git workflow, DataLad skips the staging area. A ``datalad save``
+   combines a ``git add`` and a ``git commit``, and therefore, the commit message
+   is specified with ``datalad save``.
```
mih: FTR: The reason for not using "staging" are limitations of some git-annex modes. Moreover, I personally think that it is a needless complication that is better approached by extensive commiting, and post-hoc history tuning (if desired).

docs/basics/101-2-populate.rst
```diff
@@ -0,0 +1,213 @@
...
+saving the files to the datasets history with the ``datalad save`` command.
+This time, its your turn to specify a helpful :term:`commit message`
+with the ``-m`` option:
+
+.. runrecord:: _examples/DL-101-25
+   :language: console
+   :workdir: dl-101
+   :realcommand: cd DataLad-101 && datalad save -m "add books on Python and Unix to read later"
+
+   $ datalad save -m "add books on Python and Unix to read later"
```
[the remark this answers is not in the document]
mih: Now saw that this is done further down.

docs/basics/101-3-modify.rst
```diff
@@ -0,0 +1,93 @@
...
+a DataLad dataset from scratch. Below is one way to do it, using the
+``echo`` command that comes with any Unix system, and a simple redirection
+with ``>``.
+
+.. runrecord:: _examples/DL-101-40
+   :language: console
+   :workdir: dl-101
+   :realcommand: cd DataLad-101 && echo "One can create a new dataset with 'datalad create [--description] PATH'. The dataset is created empty" > notes.txt
+
+   $ echo "One can create a new dataset with 'datalad create [--description] PATH'. The dataset is created empty" > notes.txt
```
mih: Maybe instead

```
cat << EOT > notes.txt
One can create a new dataset with 'datalad create [--description] PATH'.
The dataset is created empty
EOT
```

Better formatting, less quoting. Just as magic ;-)

adswa: I tried this, I think it doesn't work as good as echo. Even if I omit the leading ``$``, copy-pasting from the code-block to the terminal seems to copy some suboptimal formatting, and the command does not succeed:

```
cat << EOT > notes.txt
One can create a new dataset with 'datalad create [--description] PATH'.
The dataset is created empty
EOT.
heredoc else> 
```

that might not be something we want people to end up in.

adswa: sorry, my bad - nevermind.

docs/basics/101-3-modify.rst
```diff
@@ -0,0 +1,105 @@
...
+   :language: console
+   :workdir: dl-101
+   :realcommand: cd DataLad-101 && datalad save -m "Add notes on datalad create" notes.txt
+
+   $ datalad save -m "Add notes on datalad create" notes.txt
+
+Modify this file by adding another note. After all, you already know how use
+``datalad save``, so write a short summary on that as well.
+
+Again, the example below uses Unix commands (``echo`` and redirecting, but *appending*
```
mih: s/echo/cat

docs/basics/101-3-modify.rst
```diff
@@ -0,0 +1,105 @@
...
+with ``>>``) to accomplish this, but you can take any editor of your choice.
```
mih: ATM still just `>`

docs/basics/101-4-summary.rst
```diff
@@ -0,0 +1,44 @@
+Starting from scratch: Summary
+------------------------------
+
+In the last few pages, we have discovered the basics of starting a DataLad dataset from scratch,
```
mih: s/pages/section -- makes it less dependent on the actual format that people consumed

docs/basics/101-4-summary.rst
```diff
@@ -0,0 +1,44 @@
...
+and making simple modifications *locally*:
+
+* An empty dataset can be created with the ``datalad create`` command. Its useful to add a description
+  to the dataset and use the ``-c text2git`` configuration, but we will see later why.
+  Command structure: ``datalad create --description "here is a description" -c text2git PATH``
+
+* Thanks to :term:`Git` and :term:`Git-annex`, the dataset has a history to track files and their
+  modifications. Build in Git tools (``git log``) or external tools (such as ``tig``) allow to explore
```
mih: s/build in/built-in/

docs/basics/101-1-create.rst
```diff
@@ -0,0 +1,139 @@
...
+a directory on our computer to collate the materials, assignments, and
+notes in.
+
+Since this is ``DataLad-101``, lets do it as a :term:`Datalad dataset`.
+You might associate the term "dataset" with a large spreadsheet containing
+variables and data.
+But for DataLad, as we learned in the first lecture, a dataset is the core data type.
+
+As noted in :ref:`philo`, a dataset is a collection of *files*
+and *folders*, and a file is the smallest unit any dataset can contain.
```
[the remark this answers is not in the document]
adswa: this is addressed now.

docs/basics/101-1-create.rst
```diff
@@ -0,0 +1,139 @@
...
+Note the command structure of ``datalad create`` (optional bits are enclosed in ``[ ]``):
+
+``datalad create [--description "..."] [-c <config options>] PATH``
+
+
+.. runrecord:: _examples/DL-101_1
+   :language: console
+   :workdir: dl-101
+
+   $ datalad create --description "course on DataLad-101 on my private Laptop" -c text2git DataLad-101
```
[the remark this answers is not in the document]
adswa: I tried to disambiguate/clarify.

docs/basics/101-1-create.rst
```diff
@@ -0,0 +1,139 @@
...
+Find a nice place on your computers file system to put a dataset for ``DataLad-101``,
+and create a fresh, empty dataset with the ``datalad create`` command.
+In a bit of time, you will thank yourself for adding a description to
+your dataset with the *optional* ``--description`` flag. (At the moment,
+we will not dive into the details of where that description ends up or
+becomes useful in the future, but in more advanced parts of the book
+we will see how this gets handy.)
+Note the command structure of ``datalad create`` (optional bits are enclosed in ``[ ]``):
+
+``datalad create [--description "..."] [-c <config options>] PATH``
```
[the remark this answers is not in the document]
adswa: tried to distinguish that better.

docs/basics/101-1-create.rst
```diff
@@ -0,0 +1,139 @@
...
+Each of them is identified by a unique 40 character sequence, called a
+:term:`checksum`.
+Highlighted in this output is information about the author and about
+the time, as well as a :term:`commit message` that summarizes the
+performed action concisely. In this case, both commit messages were written by
+DataLad itself. The most recent change is on the top. The first commit
+written to the history therefore states that a new dataset was created,
+and the second commit to the history is related to ``-c text2git`` which
+configures :term:`Git-annex` to store textfiles in Git (but more on this
+configuration later).
```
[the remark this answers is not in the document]
adswa: Agreed - done.

docs/basics/101-2-populate.rst
```diff
@@ -0,0 +1,213 @@
...
+
+    - here you can find more guidelines: https://gist.github.com/robertpainsi/b632364184e70900af4ab688decf6f53
+
+    **DON'Ts**
+
+    - passive voice is hard to read afterwards
+
+    - extensive formatting (hashes, asterisks, quotes, ...) will most likely make your shell complain
+
+    - it should be obvious: do not say nasty things about other people
```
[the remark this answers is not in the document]
adswa: good to know, thx

docs/basics/101-3-modify.rst
adswa: this is done.

docs/basics/101-3-modify.rst
adswa: yes, thx!

docs/basics/101-3-modify.rst
adswa: thx

docs/basics/101-4-summary.rst
adswa: true, done.

docs/basics/101-4-summary.rst
adswa: thx

adswa:
I will merge this once tests pass and we can turn back to this if further developments/sections require changes.

PR #56, opened 2019-07-09
Basic DataLad magic

adswa:
I've started to work on a section on ``datalad install``. Simultaneously, I have reworked everything in the "Datasets" section that is not yet in "Starting from scratch" into this new section. This way, we can get rid of the standalone dataset section, and all of the "What is a dataset" content is brought to the reader more subtly and without the neuroimaging specific dataset.

We should discuss whether subtly introducing datasets is better than a standalone.

In these two new added sectionsy, I'm showing the basic ``datalad install`` within the ``DataLad-101`` example by installing the DataLad machinelearning-books dataset inside of the ``DataLad-101`` dataset, and afterwards I explain the concept of nesting that has just happened.


docs/basics/101-5-install.rst
```diff
@@ -0,0 +1,204 @@
...
+commands we discovered.
+
+But we've been told that with DataLad we could very easily get vast amounts of data to our
+computer. Rumor has it that this would be only a single command in the terminal!
+Therefore, everyone in todays lecture exitedly await todays topic: Installing datasets.
+
+"With DataLad, users can *install* existing
+DataLad datasets from paths, urls, or open-data collections." our lecturer begins.
+"This makes accessing data fast and easy. A dataset that others could install can be
+created by anyone, without a need for external software. Your own Datasets can be
```
mih: s/external/additional

docs/basics/101-5-install.rst
```diff
@@ -0,0 +1,204 @@
...
+(``--dataset``) flag. This specifies the dataset to perform the operation on. Because we are inside
+of the ``DataLad-101`` dataset, the pointer to the dataset is a ``.`` (which is Unix
+way for saying "current directory"). The dataset to be installed lives on Github, and
+we can give its Github URL as a source (``-s``, ``--source``).
+
+.. runrecord:: _examples/DL-101-5-1
+   :language: console
+   :workdir: dl-101/DataLad-101
+
+   $  datalad install -d . -s https://github.com/datalad-datasets/machinelearning-books.git books/ml-books
```
mih: I am torn re the use of short options. They keep things at a comprehensible length, but they are also harder to parse for people that see them for the first time. Maybe it is better to introduce them like this:
```
datalad install --dataset . --source https://github.com/datalad-datasets/machinelearning-books.git books/ml-books

# or shorter
#datalad install -d . -s https://github.com/datalad-datasets/machinelearning-books.git books/ml-books
```
in the same box. With the comments it may be clear enough that only one of the lines needs to be executed.

docs/basics/101-5-install.rst
```diff
@@ -0,0 +1,204 @@
...
+Additionally, there is another directory ``ml-books`` which contains 10 more books,
+and a ``README.md`` file.
+You might be confused by the small arrows and cryptic paths that follow every PDF -
+we'll get to what that is in the next section.
+
+
+Dataset content identity and availability information
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+You might have been surprised by how fast the datalad was installed. Shouldn't
```
mih: s/datalad/dataset

docs/basics/101-5-install.rst
```diff
@@ -0,0 +1,204 @@
...
+   $ du -sh      # Unix command to show size of contents
+
+This is tiny. So lets see whats in this dataset.
+
+.. runrecord:: _examples/DL-101-5-4
+   :language: console
+   :workdir: dl-101/DataLad-101/books/ml-books
+
+   $ ls
+
```
mih: This might be the place to introduce `datalad status --annex basic`, but https://github.com/datalad/datalad/issues/3519 :(

docs/basics/101-5-install.rst
```diff
@@ -0,0 +1,204 @@
...
+   :language: console
+   :workdir: dl-101/DataLad-101/books/ml-books
+
+   $ datalad get .
+
+Wasn't that easy?
+
+Note that any data that is already retrieved (one book) is not downloaded again.
+Datalad summarizes the outcome of the execution of ``get`` in the end and informs
+that one file was ``notneeded`` and the retrieval of the other files was ``ok``.
```
mih: that *download of* one file

docs/basics/101-5-install.rst
```diff
@@ -0,0 +1,204 @@
...
+
+You have now experienced how easy it is to share data with DataLad.
```
mih: To be fair: how easy it is to "obtain shared data"

docs/basics/101-6-nesting.rst
```diff
@@ -0,0 +1,87 @@
...
+
+We can see that it is the most recent commit hash of the subdataset
+(albeit we can see only the first seven characters here -- a ``git log``
+would show you the full hash).
+This is what is meant by "the top-level DataLad dataset (the *superdataset*) only stores
+*which version* of the subdataset is currently used".
+
+Importantly, once we learn how to make use of the history of a dataset,
+we can set subdatasets to previous states, or *update* them.
+Such an approach is especially useful for reproducible analyses.
```
mih: That is a big leap from a very basic example to reproducibility. I'd either add a little more text to make clear that this is something that is relevant, e.g. in a scientific context, and why. But it might also be fine not to bring it up, just yet.

docs/basics/101-6-nesting.rst
```diff
@@ -0,0 +1,87 @@
...
+If one uses external data, this data can be installed as a subdataset.
+Should the data get extended or changed, or exist in different versions,
+its subdataset can be updated to include the changes easily, or set to the
+state in history which contains the required version of the data.
+All of this is done without "polluting" the superdatasets history with
+potentially hundrets of file changes, but a simple subproject commit change.
+More detailed examples of this can be found in later parts of the Basics
+or in the use cases in the last
+section (for example in :ref:`remodnav`).
```
mih: Another point to make re nesting (here or maybe later) is the modular re-use of dataset in different contexts/superdatasets. This is possible and simple, precisely because all this information is kept within a dataset.

docs/basics/101-5-install.rst
adswa: oh this is such a cool command! :(
We could however switch datasets to the longnow podcast recordings? As far as I can see, this dataset does not have URL-type annex keys. The ML books weren't 100% fitting into the narrative, so we can equally well just pretend that the podcasts are audio recordings of a past years course (or something similar). What do you think?

mih: That would be great! I think that dataset is a good fit for an educational narrative. We could later on even reveal that it is auto-generated by monitoring a podcast feed:

```
git annex importfeed --template '${feedtitle}/${itempubdate}__${itemtitle}${extension}' --force http://longnow.org/projects/seminars/SALT.xml
```

Which would be a straightforward connection to a potential future chapter on getting non-DataLad data sources into the DataLad world (scraping, crawling, ...)

adswa: Very good idea, I like this! I'll rewrite this then.

mih: Cool. I have just pushed an update of this dataset. I have used `datalad run` this time. So there is now the possibility to have readers `rerun` the update and obtain any new episodes....

adswa: awesome!!!

adswa:
FTR: I'm planning to redo the graphic thats currently still in there. Its too neuroimaging to be understandable, but its a wonderful way to visualize nesting. We can use the original graphic later, and readers are already used to figures of this type.

adswa:
Note: I'll need to update the examples again once I'm not having shitty ICE WIFI anymore. One download failed (because of the shitty WIFI) and polluted some of the output.

adswa:
Does anyone have strong feelings for the current standalone [datasets section](http://handbook.datalad.org/en/latest/basics/datasets.html)? With the exception of the scientific use case, everything mentioned in the standalone section is now subtly introduced in sections ``101-1-create`` - ``101-6-nesting``. 
Seeing the current book structure as it is now I am more in favor of readers exploring what a dataset is step-by-step first. To not lose the emphasis on the key role of datasets, we could have a "recap" that summarizes what a dataset is after section ``101-7-summary`` (or later): a section just as the current standalone, but using the ``DataLad-101`` dataset in the state after ``101-7-summary`` (instead of studyforrest), and merely reminding of instead of executing commands. 
@loj @mih

loj:
> [adswa wrote]
> To not lose the emphasis on the key role of datasets, we could have a "recap" that summarizes what a 
> dataset is after section 101-7-summary (or later): a section just as the current standalone, but using 
> the DataLad-101 dataset in the state after 101-7-summary (or later)(instead of studyforrest), and 
> merely reminding of instead of executing commands.

I think this makes a lot of sense.

mih:
> current standalone datasets section

I think it can go. It does not match the style of the other parts of the book, and I personally prefer that one over "front-loading" particular not-yet-well-connected conceptual information. If we think we need a high-level conceptual overview, we can put it further back in the book. But TBH, this rather goes into an academic paper that could be referenced.

mih:
Had another read -- cool! Another big step forward. Thx!

PR #62, opened 2019-07-15
ENH: have sidebar fixed

adswa:
This prevents scrolling back to the top to switch chapters.
Addresses #61.


mih:
Ah, great find! I was about to invent a solution. Much better this way.

PR #67, opened 2019-07-16
[WIP] Managing the Filesystem in dataset

adswa:
This is very much work in progress, and more of an outline and call for additions than a PR.

Throughout learning datalad I was uncertain what certain file system operations (moving/renaming files, directories, subdatasets, the full dataset, ...) would do to my data, and in the past I also managed to spectacularly mess up dataset statuses and contents by using certain commands naively.
This section should give readers definite answers and a playground to experience file management in datasets.

Currently its only a couple of questions I collected myself. If anyone has additions, please add them. Whenever it fits, I'll also have to query @mih on a couple of details I am currently unsure about.

But this is not urgent; nothing is dependent on this, and this section will be able to live pretty much anywhere within ``Basics``. Naming (101-114) is arbitrary, and will likely change at some point.

### DIFF
--- a/docs/basics/101-114-filesystem.rst
+++ b/docs/basics/101-114-filesystem.rst
@@ -1,0 +1,127 @@
+Miscallaneous file system operations
+------------------------------------
+
+With all of the information about symlinks and object trees,
+you might be reluctant to perform usual file system managing
+operations, such as copying, moving, or renaming a files or
+directories.
+
+If I renamed on of those books, would the symlink that points
+to the file content still be correct? What happens if I'd copy
+an annexed file?
+If I moved the whole ``books/`` directory? What if I moved
+all of ``DataLad-101`` into a different place on my computer?
+What if renamed the whole superdataset?
+And how do I remove a file, or directory, or subdataset?
+
+There is no better way of learning than doing. Here, in the
+safe space of the ``DataLad-101`` course, you can try out all
+of the things you would be unsure about or reluctant to try
+on the dataset that contains your own, valuable data.
+
+What happens if I rename an annexed file?
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Let's try it. In Unix, renaming a file is exactly the same as
+moving a file.
+
+.. runrecord:: _examples/DL-101-114-101
+   :language: console
+   :workdir: dl-101/DataLad-101
+
+   cd books/
+   mv TLCL.pdf The_Linux_Command_Line.pdf
+   ls -lah
+
+Try to open the renamed file, e.g. with
+``evince The_Linux_Command_Line.pdf``.
+This works!
+
+But let's see what changed in the dataset with this operation:
+
+.. runrecord:: _examples/DL-101-114-102
+   :language: console
+   :workdir: dl-101/DataLad-101/books
+
+   $ datalad status
+
+We can see that the old file is marked as ``deleted``, and
+simultaneously, an ``untracked`` file appears: the renamed
+PDF.
+
+In theory, we would need to do a ``datalad save`` at this point
+to commit the "deletion" and the "new file". This however, is
+suboptimal, and Git has built-in commands that provide a better
+solution.
+
+Lets revert the renaming of the the files:
+
+.. runrecord:: _examples/DL-101-114-103
+   :language: console
+   :workdir: dl-101/DataLad-101/books
+
+   $ git reset --hard master       # this restores TLCL.pdf
+   $ rm The_Linux_Command_Line.pdf       # this removes the untracked copy
+
+A more convenient way to rename files is the ``git mv`` command:
+
+.. runrecord:: _examples/DL-101-114-104
+   :language: console
+   :workdir: dl-101/DataLad-101/books
+
+   $ git mv TLCL.pdf The_Linux_Command_Line.pdf
+
+.. runrecord:: _examples/DL-101-114-105
+   :language: console
+   :workdir: dl-101/DataLad-101/books
+
+   $ datalad status
+
+We can see that the old file is still seen as "deleted", but the "new",
+renamed file is "added". A ``git status`` displays the change
+in the dataset a bit more accurate:
+
+.. runrecord:: _examples/DL-101-114-106
+   :language: console
+   :workdir: dl-101/DataLad-101/books
+
+   $ git status
+
+.. todo::
+
+   can I mitigate a ``git commit`` here? I believe the best ``datalad save``
+   command to save the renaming (without the rest of the dataset state) is
+   ``datalad save -m "renamed TLCL to verbose title" The_Linux_Command_Line.pdf TLCL.pdf``,
+   but this still requires listing all paths, and tab completion does not seem to work for
+   the "deleted" files...
+
+TODO
+
+What happens if I rename a directory or subdataset?
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+What happens if I move files, directories, or subdatasets?
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The first question is easy. Because in Unix renaming is the same as moving,
+moving works exactly like renaming in the examples with the PDFs described above.
+Make sure to use ``git mv`` instead of ``mv``.
+
+TODO: directory (all files in a directory will be renamed)
+TODO: subdataset (tricky: change in .gitmodules. Also, I failed trying to
+revert a ``git mv`` with ``git reset --hard master``. It did not move
+subdataset back into original place, the subds became an untracked directory.
+
+What if I move the superdataset into a different place on my computer?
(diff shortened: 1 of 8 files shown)


docs/basics/101-114-filesystem.rst
```diff
@@ -0,0 +1,127 @@
...
+
+   $ git status
+
+.. todo::
+
+   can I mitigate a ``git commit`` here? I believe the best ``datalad save``
+   command to save the renaming (without the rest of the dataset state) is
+   ``datalad save -m "renamed TLCL to verbose title" The_Linux_Command_Line.pdf TLCL.pdf``,
+   but this still requires listing all paths, and tab completion does not seem to work for
+   the "deleted" files...
```
mih: No, after a `git mv`, a `git commit` is the most elegant way. However, there is no need for `git mv`, as Git itself doesn't really record the "move". Instead, it tends to infer things post-hoc from the recorded commits. Watch this:

```
$ datalad create demo
$ cd demo
$ echo 123 > one
$ datalad save
# plain move, not Git
$ mv one two
$ datalad save
$ git log --patch HEAD~1..HEAD
commit 4c39756a94928a529d9000b573a10ddf75991c01 (HEAD -> master)
Author: Michael Hanke <michael.hanke@gmail.com>
Date:   Tue Jul 16 13:01:01 2019 +0200

    [DATALAD] Recorded changes

diff --git a/one b/two
similarity index 100%
rename from one
rename to two
```

docs/basics/101-114-filesystem.rst
```diff
@@ -0,0 +1,127 @@
...
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The first question is easy. Because in Unix renaming is the same as moving,
+moving works exactly like renaming in the examples with the PDFs described above.
+Make sure to use ``git mv`` instead of ``mv``.
+
+TODO: directory (all files in a directory will be renamed)
+TODO: subdataset (tricky: change in .gitmodules. Also, I failed trying to
+revert a ``git mv`` with ``git reset --hard master``. It did not move
+subdataset back into original place, the subds became an untracked directory.
```
mih: Without a `datalad move` this is all a mine field, see e.g. https://github.com/datalad/datalad/issues/3464

docs/basics/101-114-filesystem.rst
```diff
@@ -0,0 +1,127 @@
...
+
+What if I move the superdataset into a different place on my computer?
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+TODO: I don't think it should be a problem, but not sure whether I'm
+missing something
```
mih: In standard situation this should be OK.

docs/basics/101-114-filesystem.rst
```diff
@@ -0,0 +1,127 @@
...
+
+What happens if I copy files?
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+TODO: Copying will create a new file that could be added. Its not symlinked,
+I don't yet understand why, but the file is write-protected (why exactly? bc
+symlinked content is write protected?), which is
+inconveniently difficult to explain (needs digression into permissions)
```
mih: Continuing the example from above:

```
$ cp two copyoftwo
$ datalad save
```

should work just fine. On save, git-annex will automatically identify any identical files and symlink them to the same annex object.

docs/basics/101-114-filesystem.rst
mih: It is a bit trickier to `mv` a file into a subdirectory:

```
$ mkdir subdir
$ mv two subdir/two
```

The symlink will be broken. In any case, a `datalad save` also rectifies that.

In contrast to the above, Git will not detect the rename. Why? Because the content that is tracked in Git is the actual symlink, and due to the change in location, the nature of the symlink has to change. Hence, what looks like a move on the filesystem is actually a move plus a content change for Git.

docs/basics/101-126-filesystem.rst
```diff
@@ -0,0 +1,422 @@
...
+You can use standard Unix commands such as ``mv`` to do it,
+and also which ever graphical user interface or explorer you may
+use.
+
+Beware of one thing though: If your dataset either is a sibling
+or has a sibling with the source being a path, moving or renaming
+the dataset will break the linkage between the datasets. This can
+be fixed easily though. Let's try it.
+
+As section (todo: link section on config files) explains, each
```
adswa: Given our conversation last week, I'm not sure whether this is a best practice.
TL;DR: Lets say I rename/move a dataset that is another datasets sibling -- is it okay to adjust the remote url in the .git/config file? It works, but is there a preferred way?

adswa:
I think I have basic answers to some common questions now. I appreciate feedback on those, and also any other ideas about common questions that could arise.

adswa:
Note to self:
Other helpful operations might be:

- "deleting" (uninstalling) subdatasets
- deleting a superdataset (this will digress a bit into permissions)
- "deleting" (dropping) annexed content

adswa:
I think I have basic answers to some common questions now. I appreciate feedback on those, and also any other ideas about common questions that could arise.

docs/basics/101-140-filesystem.rst
```diff
@@ -0,0 +1,764 @@
...
+      :workdir: dl-101/DataLad-101/books
+
+      $ git log -1 -p
+
+   As you can see, this action does not show up as a move, but instead
+   a deletion and addition of a new file. Why? Because the content
+   that is tracked is the actual symlink, and due to the change in
+   relative location, the symlink needed to change. Hence, what looks
+   and feels like a move on the file system for you is actually a
+   move plus a content change for Git.
```
mih: A pointer to the `git annex fix` command could be added to this box -- which is the one that is triggered by `datalad save` -> `git commit`, and actually does the job of fixing things up inside.

docs/basics/101-140-filesystem.rst
```diff
@@ -0,0 +1,764 @@
...
+one has to regain write permissions to everything in the dataset. This is done
+with the `chmod <https://en.wikipedia.org/wiki/Chmod>`_ command::
+
+    chmod -R u+w <dataset>
+
+This *recursively* (``-R``, i.e. throughout all files and (sub)directories) gives users
+(``u``) write permissions (``+w``) for the dataset.
+
+Afterwards, ``rm -rf <dataset>`` will succeed. Be aware though that this will
+irretrievable delete the dataset, it's contents, and it's history.
```
mih: `datalad remove <superdataset>` (likely with `--nocheck` to be faster, and possibly with `--recursive` in case of subdatasets) is the dedicated DataLad command to achieve this. However, one (or at least me) often performs an `rm -rf` first, sees the error shown above, regrets doing so, and has to fix up exactly in this way -- hence having this info here is crucial.

Just as `rm -rf .` will not work, `datalad remove .` is not possible -- removal only works from a working directory outside the dataset.

docs/basics/101-140-filesystem.rst
```diff
@@ -0,0 +1,764 @@
...
+
+Summary
+^^^^^^^
+
+To sum up, the most filesystem management operations are safe, and easy.
```
mih: s/most//

docs/basics/101-140-filesystem.rst
```diff
@@ -0,0 +1,764 @@
...
+
+To uninstall the dataset, use
+
+.. runrecord:: _examples/DL-101-140-161
+   :language: console
+   :workdir: dl-101/DataLad-101
+
+   $ datalad uninstall cloud
+
+Note that the dataset is still known in the dataset, and not completely removed.
```
bpoldrack: I guess, this might need a sentence or two, i.e. what is deleted in what sense and what is not.
May be just point out, that you can get it again in the exact same state as before via `datalad install` (or `datalad get`), which is not true, if it's "really" removed from the dataset.

adswa:
Thanks a lot, @kyleam and @bpoldrack, these remarks were very helpful!

docs/basics/101-140-filesystem.rst
```diff
@@ -0,0 +1,813 @@
...
+If I renamed one of those books, would the symlink that points
+to the file content still be correct? What happens if I'd copy
+an annexed file?
+If I moved the whole ``books/`` directory? What if I moved
+all of ``DataLad-101`` into a different place on my computer?
+What if renamed the whole superdataset?
+And how do I remove a file, or directory, or subdataset?
+
+Therefore, there is an extra tutorial offered by the courses'
+TA today, and you attend.
```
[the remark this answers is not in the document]
adswa: The book has a story-line to not make everyone feel like they are reading the documentation  ;-) (Zebava even promised to create some illustrations!)

docs/basics/101-140-filesystem.rst
```diff
@@ -0,0 +1,813 @@
...
+      :language: console
+      :workdir: dl-101/DataLad-101
+
+      $ ls -l copyofTLCL.pdf
+      $ ls -l books/TLCL.pdf
+
+   Indeed! Apart from their relative location (``.git`` versus
+   ``../.git``) their symlink is identical. Thus, even though two
+   copies of the book exist in your dataset, your disk needs to
+   store it only once.
```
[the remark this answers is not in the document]
adswa: oh, thats a good point...thx

PR #72, opened 2019-07-17
Sharing on a local file system

adswa:
Adding some work in progress on some content on

- installing a Datalad from a path (instead of URL)
- showcasing some principles of ``datalad install`` with and without ``--recursive``
- introducing ``git annex whereis`` and starting to talk about where file content can be retrieved from
- simple ``datalad update --merge``

Will need to be integrated with the rest of current book and wrapped into the narrative, among other TODOs.


mih:
Nice, thx! Pushed a few tweaks. Good to go!

adswa:
Thx!

Issue #81, opened 2019-07-21
Cross-reference technical docs


mih:
#79 reminded me that it would be good to have a standard way to cross-reference the book content with the rest of the technical docs. For example, every command has a "manpage" that can be found via this pattern (here for `create`):

http://docs.datalad.org/generated/man/datalad-create.html

and a corresponding page with analog information for Python API users:

http://docs.datalad.org/generated/datalad.api.create.html

ATM when this command is introduced in http://handbook.datalad.org/basics/101-101-create.html no reference to this additional documentation is made.

I see at least to possible approaches:

1. we make an immediate reference whenever a command first appears (in some kind of a note)

2. we maintain a proper list of all described commands and their functionality (much like #79 id doing) and we use proper index markup (http://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html?highlight=index%20generating#index-generating-markup) to cross-ref this list with the book content, and only place the cross-ref to other technical docs in this list.

I haven't played with (2) yet, but I somewhat feel that it would be best to do this in a single place, rather than all over the book.

adswa:
I think this is a good idea. Maybe I haven't found the right parts of the docs (whether there is an easier way), but a possible solution that would fulfill 2. is to populate a ``Datalad Command Reference`` document with a second ``.. glossary::`` directive. This second glossary can contain links to the external docs. Within the first occurrence of a command we can provide an ``.. index::`` entry. Thus, the index that is automatically generated contains two references, one to the book content, one to the glossary. The glossary in turn references the docs.

I've push a minimal example as a PR in #82.

mih:
We are now aiming for a link to the Python API reference at http://docs.datalad.org/modref.html when there is a chapter on the difference of cmdline and Python API.

Otherwise, a link to the command manpage is include at the place where a command is mention first. Additionally, this location is marked up in the book's index as the primary index item for this command.

An example is already in master: 2db723b7d7ba32981eca109a24c057dd5b46f1eb

PR #82, opened 2019-07-21
index references

adswa:
This is a minimal example for a potential implementation of the idea brought up by @mih in #81.
If there are easier or different ways to do it, please let me know.


mih:
This could work. I am a bit concerned whether we are needlessly boxing ourselves into the glossary markup. Maybe a plain table with the covered commands and links to further info is more readable. But ATM I don't have a good idea yet, what information we might want to gather in such a structure.

adswa:
Now we have all of the commands in this structure.

adswa:
updated index building

PR #84, opened 2019-07-22
About gitignore files

adswa:
This section is on ``.gitignore`` files, and aims to educate people on how to keep unwanted content out of version control by using such files, and also mentions the ``datalad save -u`` option again.


docs/basics/101-130-ignore.rst
```diff
@@ -0,0 +1,74 @@
...
+will save the complete datasets status if one does not provide
+a path to a file change. This would result in all content
+that is either modified or untracked being saved in a single
+commit.
+This is the reason why a ``datalad run`` requires a clean
+dataset: The ``datalad save`` that a ``datalad run`` ends with
+internally should only save changes that can be attributed to
+the command that was run, and not changes that existed already
+but were yet unsaved.
+Alternatively, you learned that one can specify the ``--explicit``
```
mih: See my thoughts on this in #88, I think this "it-is-OK-to-be-messy" idea should appear as late as anyhow possible.

docs/basics/101-130-ignore.rst
```diff
@@ -0,0 +1,74 @@
...
+``playlists/my-little-pony-themesongs/*``.
+
+To use a ``.gitignore`` file, simply create a file with this
+name in the root of your dataset (be mindful: remember the leading ``.``!).
+You can use one of `thousands of publicly shared examples <https://github.com/github/gitignore>`_,
+or create your own one.
+
+You just need to save it once to your dataset and then you're good.
+If you have new content you don't want to track, you can add
+new paths or patterns to the file.
```
mih: A thing to remember for later: If a command creates an output that is git-ignored, a subsequent command that requires it as an undisclosed input will succeed, if both commands a ran in succession. However, the second command will fail, if re-ran on its own.

adswa:
I've revisited this after a while, but I am uncertain whether it should be included or not. IMO it could fit into a yet to come, completely optional, and very late occurring section on Git (which would include an overview of common operations to play around with the history, and maybe an explanation of the concepts of HEAD, branches, refs, ...). But I would also be completely fine leaving this section on ``.gitignore`` out completely.

mih:
I think this is a good thing to have. I personally always have stuff in a dataset that I need for the moment, but I do not want to track. We either have to explain it, or we make it unnecessary with some convention and pre-configuration. E.g. we could have the `yoda` config procedure put a top-level `tmp/` directory into a default `.gitignore` file.

Maybe we can assemble a chapter on "Being messy -- with style" and put this in there.

adswa:
Sounds good.

mih:
Thx, will merge

Issue #88, opened 2019-07-23
Construct all examples to always use `datalad save` without path args


mih:
This should communicate that one can employ a "clean-desk" philosophy when working with datasets. A (later) dedicated section can detail how one can also survive in a messy world, and when this might be necessary. But otherwise it just complicates the world -- for now good reason, IMHO.

mih:
Done.

PR #89, opened 2019-07-24
de-emphasize path specs in dl save

adswa:
Addressing #88, I'm starting to de-emphasize path specifications in ``datalad save``.


adswa:
I have removed all superfluous path specifications to address #88.

mih:
Great, thx!

PR #106, opened 2019-08-06
Git-heavy section on interactions with the history

adswa:
Even though we tried to stay completely away from Git, I believe it is not possible to exclude it. 
The dataset features of reverting mistakes or viewing past versions of files are a major magic-factor and very intriguing to users (especially if they haven't used Git before) - at least to me the most intriguing feature was that I could "go back in time".
I also anticipate that certain errors (e.g. accidentally saving things that should not be saved (together)) are common, and among some general Git magic, this should show readers a way out of these problems. 

My proposal is to put this section into the chapter on "Getting help". It gives a glimpse into what is generally possible and a safe space to explore, but its not a comprehensive guide. Instead it has some pointers to external resources (and also sufficient warnings to keep users wary of using such operations without thought).


adswa:
It took me a while to write this section - it is now a glimpse into ``git commit --amend``, ``git reset``, ``git checkout`` and interactive rebasing, and a safe-space to explore those commands worry-free. I'd appreciate critical feedback!

mih:
I haven't read the chapter yet, but before I forget, here is a pointer to a desperately needed command that is also relevant in this conext: https://github.com/datalad/datalad/issues/1424

adswa:
thx for the pointer. I agree, this would be relevant indeed.

docs/basics/101-136-history.rst
```diff
@@ -0,0 +1,767 @@
...
+   :workdir: dl-101/DataLad-101
+   :realcommand: echo "$ git checkout $(git rev-parse HEAD~15)" && git checkout $(git rev-parse HEAD~15)
+
+How did your ``notes.txt`` file look at this point?
+
+.. runrecord:: _examples/DL-101-136-123
+   :language: console
+   :workdir: dl-101/DataLad-101
+
+   $ cat notes.txt
```
mih: Maybe introduce `cat-file`?

docs/basics/101-136-history.rst
```diff
@@ -0,0 +1,767 @@
...
+   :workdir: dl-101/DataLad-101
+
+   $ datalad save -m "add joke evaluation to joke" Gitjoke2.txt
+
+How could this modification to ``Gitjoke2.txt`` be undone?
+With the :command:`git reset` command again. If you want to
+"unsave" the modification but keep it in the file, use
+:command:`git reset --mixed` as before. However, if you want to
+get rid of the modifications entirely, use the option ``--hard``
+instead of ``--mixed``:
```
mih: Maybe point out that it only restores just the one file, because only that one file was modified. If the commit would have had other modifications, all of them would have been undone. This is one of the reasons why one should strive for commits to represent meaningful logical units of change.

docs/basics/101-136-history.rst
```diff
@@ -0,0 +1,767 @@
...
+   :language: console
+   :workdir: dl-101/DataLad-101
+
+   $ datalad save -m "add note on helpful git resource" notes.txt
+
+The objective is to remove the first, "bad" modification, but
+keep the more recent modification of ``notes.txt``. A :command:`git reset`
+command is not convenient, because resetting would need to reset
+the most recent, "good" modification as well. Instead, an interactive
+rebase with the command :command:`git rebase -i` is necessary [#f5]_.
```
mih: The presented approach is valid, but also problematic in any collaborative scenario where the respective branch was pushed already, and others may have already built on on. Rebasing in this case leads to a world of pain that quickly escalates and requires everyone involved to have some expert level Git foo.

The most compatible, least intrusive way revert one or more changes is `git revert` IMHO. It will produce (an) additional commit(s) (instead of rewriting history). This makes it obvious in the history that a mistake happened that was rectified. Any consumer can simply merge this change and also sees that a mistake has been fixes -- in contrast to a rather obscure merge conflict.

Most of the text on conflict resolution below remains valid in the context of `git revert` and can and should be kept.

docs/basics/101-136-history.rst
```diff
@@ -0,0 +1,767 @@
...
+
+   The :command:`git commit --amend` commands will let you
+   rewrite the commit message of the most recent commit. If you
+   however need to rewrite commit messages of older commits, you
+   can do so during a so-called "interactive rebase". The command
+   for this is
+
+   .. code-block:: bash
+
+      $ git rebase -i HEAD~N
```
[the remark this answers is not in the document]
adswa: There is a footnote about it:

```rst
.. [#f4] Note though that rewriting history can be dangerous, and you should
         be aware of what you are doing. For example, rewriting parts of the
         dataset's history that have been published (e.g., to a Github repository)
         already or that other people have copies of, is not advised.

```

I will keep the footnote, but also include an explicit statement in the text.

docs/basics/101-136-history.rst
adswa: very good point, thx.

docs/basics/101-136-history.rst
adswa: Done.

docs/basics/101-136-history.rst
```diff
@@ -0,0 +1,767 @@
...
+
+As expected, all of the modifications present prior to the
+command are saved into the most recent commit, and the commit
+message DataLad provides by default, ``[DATALAD] Recorded changes``,
+is not very helpful.
+
+Changing the commit message of the most recent commit can be done with
+the command :command:`git commit --amend`. Running this command will open
+an editor (the default, as configured in Git), and allow you
+to change the commit message.
```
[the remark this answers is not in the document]
adswa: I'd be fine with either waiting with this PR or modifying it afterwards to reflect any changes that will be implemented.

adswa:
Cool, thx :-)

PR #107, opened 2019-08-06
add note on unix tools not preinstalled in OSX

adswa:
I gave a short Datalad intro based on the book today at the Neurohackademy. I realized that many Mac users did not by default had Unix tools my Debian system had, so I added footnotes about them.


mih:
Good point. I will merge this now, but we might want to think about an alternative of making these remarks that is a little less redundant.

PR #111, opened 2019-08-19
Sharing data on 3rd party infrastructure

### LINKED ISSUE #186: Use case: dataset hosting on GIN
See https://github.com/datalad/datalad/issues/2680

This is particularly appealing for novice users, because:

- it is free
- it also uses git-annex under the hood, hence is fully compatible with everything we do on datalad
- unlike GitHub, has a UI that resolves annex a symlinks to the real file content

adswa:
I'm creating some space for a section that should showcase how to share a dataset across more than a shared file system. Easiest to start with might be dropbox or similar, but I will yet have to read into how to do that.

This fixes #186.

TODO

- [x] integrate in narrative
- [x] add summary


mih:
The key to this is to use (in almost all cases) the appropriate git-annex special remote implementation to first push and later obtain file from a 3rd-party service. How to do that exactly very much depends on the service. For example:

### box.com/own-cloud/sciebo
They are accessible through a WEBDAV interface, which git-annex has built-in support for

```
git annex initremote box.com \
   type=webdav url=https://dav.box.com/dav/team/project_one \
   chunk=50mb encryption=none
```

3rd-party services often limit the file size -- git annex can "chunk" files arbitrarily and hence is capable to easily circumvent such constraints -- on download, files will be merged back together to provide a seamless experience.

### rclone (Google, Amazon, MS, Dropbox, ...)

The majority of commercial providers is made accessible via `rclone` (https://git-annex.branchable.com/special_remotes/rclone/; https://github.com/DanielDent/git-annex-remote-rclone)

### general workflow

Examples of how things can be set up can be found in the demos here: https://www.datalad.org/for/data-publication

The key is to realize that the remote 3rd-party storage services do not get to see the actual repository content (real filenames, directories, etc) -- those only live in the dataset/repo which can be hosted at another place or service (github/lab, ...). What is stored on dropbox and friends is merely a tree of annex objects. Only through this design it becomes possible to chunk files into smaller units, optionally encrypt content on its way from a local machine to a storage service, and avoid leakage of information via filenames. Therefore these places are not something a real person would take a look at, instead they are only meant to to be managed and accessed via datalad/git-annex (and in the case of encryption that is pretty much required). This can be a downside, as it will not be possible to pass a dropbox link to such a project to a random stranger that is not using datalad. But....

### Related

datalad  also has some support for "exporting" data to other services. For example the `export-to-figshare`. The main difference is that this moves data out of versionen und decentralized tracking, and essentially "throws it over the wall". Alternatively, git annex provides "export/input" functionality that can be used to read and write from/to a particular "human-facing" representation (which is not a git repo), for example the content of a particular version of a particular branch. 

https://git-annex.branchable.com/git-annex-export/
https://git-annex.branchable.com/git-annex-import/

adswa:
I spend the day wrangling ``rclone`` special remotes, but now I have a draft for how to publish a dataset and its annexed content with a publication dependency on Dropbox. At this point, I need to think of a way to work this into the narrative...

adswa:
...one thought: This could become a use case (its very step-by-steppy...).

adswa:
if we do #326, this needs minor changes as well.

adswa:
I think it is useful to integrate #186 into the chapter I am starting here. In the end, this can become a quite advanced chapter that introduces various ways to share datasets with annexed contents. From my own experiences and others feedback, including data sharing (beyond shared computational infrastructure) into the Basics is very useful.

adswa:
ping @mih

docs/basics/101-138-sharethirdparty.rst
```diff
@@ -0,0 +1,502 @@
...
+that shows how you can create an S3 special remote and share data with anyone who
+gets a clone of your dataset, without them needing Amazon AWS credentials. Likewise,
+the documentation provides step-by-step walk-throughs for many other services,
+such as `Google Cloud Storage <https://git-annex.branchable.com/tips/using_Google_Cloud_Storage/>`_,
+`Box.com <https://git-annex.branchable.com/tips/using_box.com_as_a_special_remote/>`__,
+`Amazon Glacier <https://git-annex.branchable.com/tips/using_Amazon_Glacier/>`_,
+`OwnCloud <https://git-annex.branchable.com/tips/owncloudannex/>`__, and many more.
+Here is the complete list: `git-annex.branchable.com/special_remotes/ <https://git-annex.branchable.com/special_remotes/>`_.
+
+For Dropbox, the relevant special-remote to configures is
```
mih: Everything that follows would benefit a lot from better support on the datalad end https://github.com/datalad/datalad/issues/1784

docs/basics/101-138-sharethirdparty.rst
```diff
@@ -0,0 +1,502 @@
...
+   |         datalad siblings -d "/Users/awagner/Documents/DataLad-101" enable -s dropbox-remote
+
+This means that someone who wants to access the data from dropbox needs to
+enable the special remote.
+For this,  this person first needs to install and configure ``rclone``
+as well: Since ``rclone`` is the protocol with which
+annexed data can be transferred from and to Dropbox, anyone who needs annexed
+data from Dropbox needs this special remote. Therefore, the first steps are
+the same as before:
+
```
mih: And because they are identical, I would not repeat them verbosely. Just like `install rclone as described before` etc.

docs/basics/101-138-sharethirdparty.rst
```diff
@@ -0,0 +1,502 @@
...
+.. rubric:: Footnotes
+
+.. [#f1] :term:`GitLab`, on the other hand, provides a git-annex configuration. It
+         is disabled by default, and to enable it you would need to have administrative
+         access to the server and client side of your GitLab instance. Find out more
+         `here <https://docs.gitlab.com/ee/administration/git_annex.html>`_.
+
+.. [#f2] If you have an unused computer and a bit of time to spare, here is an
+         instruction `how to build your own webserver <https://www.instructables.com/id/Set-up-your-very-own-Web-server/>`_.
```
mih: While technically simple, I would be reluctant to advise people who have no experience in running web infrastructure to do so after reading an online post. Soon this thing will be a spam bot and we will all have to suffer from it ;-)

docs/basics/101-139-gin.rst
```diff
@@ -0,0 +1,214 @@
+.. _gin:
+
+Dataset hosting on GIN
+----------------------
+
+`GIN <https://gin.g-node.org/G-Node/Info/wiki>`__ (G-Node infrastructure) is a
+free data management system designed for comprehensive and reproducible management
+of scientific data. It is a secure, web-based repository store and provides
+fine-grained access control to share data. :term:`GIN` builds up on :term:`Git` and
```
mih: This sounds like we have certified that. Either remove the "secure" claim, or state "it is advertised as". I prefer the former.

docs/basics/101-139-gin.rst
```diff
@@ -0,0 +1,214 @@
...
+:term:`git-annex`, and is an easy alternative to other third-party services to host
+and share your DataLad datasets [#f1]_.
+
+Prerequisites
+^^^^^^^^^^^^^
+
+In order to use GIN for hosting and sharing your datasets, you need to
+
+- register
+- upload your :term:`SSH key` for SSH access
```
mih: worth explicitly saying "public key"

docs/basics/101-139-gin.rst
```diff
@@ -0,0 +1,214 @@
...
+   and it also shows you how to add your public SSH key to your GitHub account
+   so that you can install or clone datasets or Git repositories via ``SSH`` (in addition
+   to the ``http`` protocol), and the same procedure applies to GitLab and Gin.
+
+   Don't be intimidated if you have never done this before -- it is fast and easy:
+   First, you need to create a private and a public key (an SSH key pair).
+   All this takes is a single command in the terminal. The resulting files are
+   text files that look like someone spilled alphabet soup in them, but constitute
+   a secure password procedure.
+   You keep the private key on your own machine (the system you are connecting from),
```
mih: ...and only you have access to. Anyone who get their hands on the private key, has access to anything that the public key is protecting. If the private key doesn't have a passphrase, simply copying this file grants a person access.

docs/basics/101-139-gin.rst
```diff
@@ -0,0 +1,214 @@
...
+   and copy the public key to the system or service you are connecting to.
+   On the remote system or service, you make the public key an *authorized key* to
+   allow authentication via the SSH key pair instead of your password. This
+   either takes a single command in the terminal, or a few clicks in a web interface
+   to achieve.
+   For extra security, you can protect your SSH keys on your machine with a
```
mih: I would recommend not calling it "extra security". Instead say something like "it should be protected by a passphrase to prevent..."

docs/basics/101-140-summary.rst
```diff
@@ -0,0 +1,54 @@
...
+Summary
+-------
+
+Without access to the same computational infrastructure, you can share your
+DataLad datasets with friends and collaborators by leveraging third party
+services. DataLad integrates well with a variety of free or commercial services,
+and with many available service options this gives you freedom in deciding where
+you store your data and thus who can get access.
+
+- An easy, free, and fast option is `GIN <https://gin.g-node.org>`_, a secure,
```
mih: Same issue with "secure" here.

docs/basics/101-138-sharethirdparty.rst
```diff
@@ -0,0 +1,502 @@
...
+   :emphasize-lines: 7-8, 22, 26, 30, 36
+
+   $ rclone config
+    2019/09/06 13:43:58 NOTICE: Config file "/home/adina/.config/rclone/rclone.conf" not found - using defaults
+    No remotes found - make a new one
+    n) New remote
+    s) Set configuration password
+    q) Quit config
+    n/s/q> n
+    name> dropbox-remote
```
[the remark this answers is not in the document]
adswa: More or less arbitrary, but it helped to make the link between "you need to configure a special remote" and running the ``rclone config`` command and it distinguishes the name of the remote from the name of the storage type (``dropbox``). Especially in the ``git annex initremote`` command it makes it less ambiguous whether ``target=dropbox-remote`` is the created name or an existing value for this option. I've tried it without the prefix and it works (with slightly more ambiguousness), but personally I'm not annoyed enough by the ``-remote`` prefix to sacrifice the unambigousness... I briefly thought about renaming it to "my-dropbox", but that would not make sense to someone I could be sharing with... I think I will leave that as is.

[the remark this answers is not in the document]
mih: In that case, I'd say it is still best to qualify whose dropbox this is, or what it is for. There could be any number of additional dropbox remotes configured in a real dataset. For example, one could use two dropboxes, one shared with a select few and one that is semi public (connecting to the audience aspect you bring up in the chapter). Including some kind of discrimination identified in the name would be good.

adswa:
Thanks much for the comments! I created #344 in response to your commit message in 
576f785

docs/basics/101-138-sharethirdparty.rst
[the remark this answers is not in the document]
adswa: Ok, true, I'll change it to ``dropbox-for-friends``.

adswa:
Tests passed, I will merge, and then release it into the open!

PR #112, opened 2019-08-19
Mention open neuro in FAQ

adswa:
I'm adding a placeholder for a use case with an Open Neuro dataset. This could be a short and easy "consume a domain-specific dataset" example.


adswa:
Instead of making this a usecase, I have added the information from the datalad website on openneuro datasets into the FAQ section.

adswa:
needs minor adjustments if #326 happens.

Issue #109, opened 2019-08-16
Dedicated markup for further reading


mih:
Turn constructs like the following into dedicated directives that indicate their purpose

```
.. container:: toggle
￼	
￼	   .. container:: header
￼	
￼	      **Addition: More on this...**
```￼	

Candidate name: `moreinfo`

It would be a toggle box in HTML, and possibly a marginnote in Latex.

adswa:
That would be awesome

adswa:
This was fixed by https://github.com/datalad-handbook/book/pull/120

PR #115, opened 2019-08-20
Slightly less crappy toggle containers

adswa:
This makes the "additional sections" in Sphinx containers a bit more visible. I added border and background, exchanged the unicode arrow, and adjusted existing headers to not be completely bold (as salience with color hopefully suffices).

Its by no means what we want to have for the future (#109), but in principle a bit better than what we currently have.


mih:
I promise to deliver a dedicated directive, eventually!

PR #104, opened 2019-08-06
Dataset configurations

adswa:
Very much work in progress, I'm just adding some progress on a pretty tough section on dataset configurations (``.gitconfig``, ``.git/config``, ``.gitmodules``, ``.datalad/config``, ``.gitattributes``) consisting of minor bits and tweaks I made whenever there was a bit of time. There is a lot missing still, and I also still need to understand some things myself.
It's not worth proofreading/extensively reviewing at this point.

PR #118, opened 2019-08-21
add yet missing example files

adswa:
#104 missed some example files. I don't know why the build didn't fail, to be honest. As far as I can see, there are no other example files with these names in this repository (``_examples/DL-101-125-109``  - ``_examples/DL-101-125-112``). The rendered version included really funky examples instead of failing. For example:

```
$ datalad status
bash: line 1: datalad: command not found
```

or a git diff displays the ``docs/conf.py`` contents:

```
$ git diff
diff --git a/docs/conf.py b/docs/conf.py
index 19b18d8..7890263 100644
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -310,3 +310,180 @@ plantuml_latex_output_format = 'pdf'
 
 def setup(app):
     app.add_stylesheet('custom.css')
+
+
+
+
+###########################################################################
+#          auto-created readthedocs.org specific configuration            #
+###########################################################################
+
+
+#
+# The following code was added during an automated build on readthedocs.org
+# It is auto created and injected for every build. The result is based on the
+# conf.py.tmpl file found in the readthedocs.org codebase:
+# https://github.com/rtfd/readthedocs.org/blob/master/readthedocs/doc_builder/templates/doc_builder/conf.py.tmpl
+#
+
+
+import importlib
+import sys
+import os.path
[...]
```

I'm clueless where these snippets stem from, but adding the correct example files will hopefully fix the rendered version.

### DIFF
--- a/docs/basics/_examples/DL-101-125-109
+++ b/docs/basics/_examples/DL-101-125-109
@@ -1,0 +1,1 @@
+$ git config --file=.gitmodules --replace-all submodule."recordings/longnow".url "git@github.com:datalad-datasets/longnow-podcasts.git"
--- a/docs/basics/_examples/DL-101-125-110
+++ b/docs/basics/_examples/DL-101-125-110
@@ -1,0 +1,2 @@
+$ datalad status
+ modified: .gitmodules (file)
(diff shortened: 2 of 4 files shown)


mih:
Strange indeed, need to investigate later on.... Thx for catching it.

adswa:
Yes, it's fixed now. Thx for merging!

PR #119, opened 2019-08-21
Fix github button setup

mih:


adswa:
Thx for catching this!

PR #120, opened 2019-08-22
'findoutmore' RST directive, plus exemplary use

mih:
Visual impression should be unchanged from status quo. But now we have it semantically marked-up and can distinguish it from other boxes and toggle containers.


mih:
Cool!

PR #123, opened 2019-08-24
Finalize Remodnav

adswa:
This adds the missing parts to "Writing a reproducible paper" use case.


mih:
https://github.com/datalad-handbook/book/pull/123/files#diff-97592fd554eb5c9fd20792cef0d44d16R118

and all data should live in the ``data/`` directory.
->
and all input data should live in the ``data/`` directory.

To make it clearer that inputs are subdatasets, while outputs can go anywhere in the superdataset.

docs/usecases/remodnav.rst
```diff
@@ -291,45 +290,137 @@ to fill tables with unique names with minimal work, for example like this:
      \end{tabular*}
    \end{table}
```
mih: It would probably be sufficient to trim this down to one result column and (maybe) two sections for better readabiity and formatting in the PDF version. The table screenshot could be cropped to match.

docs/usecases/remodnav.rst
```diff
@@ -291,45 +290,137 @@ to fill tables with unique names with minimal work, for example like this:
...
             kappa = cohen_kappa_score(comb[0], comb[1])
             label = 'kappa{}{}{}'.format(rating, stim, ev)
+            # print the result
             print('\\newcommand{\\%s}{%s}' % (label, '%.2f' % kappa))
 
 
-Running the python script hence will print plenty of LaTeX commands to your screen (try it out,
-if you want!). Those statements just need to be captured, and bound to the ``.tex`` file of your
-script.
+Running the python script hence will print plenty of LaTeX commands to your screen (try it out
```
mih: not sure but: "hence will" -> "will hence" ?

docs/usecases/remodnav.rst
```diff
@@ -291,45 +290,137 @@ to fill tables with unique names with minimal work, for example like this:
...
 
-Upon compilation of the ``.tex`` file into a PDF, the results of the computations captured with
-``\newcommand`` definitions are inserted into the respective part of the manuscript.
+Upon compilation of the ``.tex`` file into a PDF, the results of the
+computations captured with ``\newcommand`` definitions are inserted into the respective part
+of the manuscript.
 
+The last step is to automate this procedure. So far, the script would need to be executed
+with a command line call, and the PDF compilation would require another commandline call.
 To automate this process, `Makefiles <https://en.wikipedia.org/wiki/Make_(software)>`_ can help.
```
mih: Maybe state that `make` is a dinosaur from decades ago that is known to many, and its advantage is that it will deliver results regardless of what actually needs to be done. It is always just `make`, regardless of programming language, operating systems, etc.

However, the makefile below will look alien to people who haven't worked with this yet. Would be could to state that any kind of "top-level" script that can be executed without parameter would do the job (in some sense).

docs/usecases/remodnav.rst
```diff
@@ -291,45 +290,137 @@ to fill tables with unique names with minimal work, for example like this:
...
+Here is the one used for the manuscript:
+
+.. code-block:: make
```
mih:
```diff
 .. code-block:: make
+   :linenos:
 
    all: main.pdf
```

docs/usecases/remodnav.rst
```diff
@@ -291,45 +290,137 @@ to fill tables with unique names with minimal work, for example like this:
...
+  the target ``main.pdf`` can be made by running the command ``latexmk -pdf -g``"
+
+- Line 5-6: "To make the target ``results_def.tex``, the script ``code/mk_figuresnstats.py`` is
+  required. If the file is present, the target ``results_def.tex`` can be made by running the
+  command ``bash -c 'set -o pipefail; code/mk_figuresnstats.py -s | tee results_def.tex'``"
+
+This triggers the execution of the script, collection of results in ``results_def.tex``, and PDF
+compilation upon typing ``make``.
+The last three lines define that a ``make clean`` removes all computed files, and also all
+images.
```
mih: Please mention `datalad run make` in the conclusions. Only this final piece will fully connect the pieces and unambiguously document who the results came to be in the dataset history.

adswa:
Thanks for the in-depth feedback!

PR #125, opened 2019-08-25
Artwork pulled in from https://github.com/datalad-handbook/artwork

mih:
Confirmed that readthedocs will pull it in automatically.


mih:
Merge to than double-check readthedocs

PR #128, opened 2019-08-26
Fix bugs found on brainbfast

adswa:
Potpourri of minor edits from sequentially executing the currently rendered book start to end.


mih:
As discussed, this should eventually be replaced by some indication with directory a code snippet is meant to be executed in. The metadata itself is already in the source files, we just need to update the rendering.

mih:
Hey, @all-contributors, please add @adswa for this test effort.

PR #134, opened 2019-08-30
Use cases from the datalad docs

adswa:
This "imports" the use cases from the datalad documentation into the handbook's use case part. Those use cases fit well into the scope of this part, and are concise summaries of principles taught in the basics.


mih:
I am unsure whether we should merge this now, or spend some time on figuring out whether we can turn the static code blocks into runable code. I am worried that we will not be able to discover if the code  gets broken at some point in the future, if we don't actually execute things.

mih:
Apart from my comment above this look great. I pushed a minor edit.

adswa:
I also thought about making the code executable - since we're now two with that thought, I'll make the code properly executable! Thx!

docs/usecases/provenance_tracking.rst
```diff
@@ -0,0 +1,293 @@
...
+discard the change, using standard Git tools [#f3]_.
+
+.. code-block:: bash
+
+   $ git reset --hard HEAD~1
+     HEAD is not at 643c175 [DATALAD RUNCMD] blur image
+
+He knows that within a DataLad dataset, he can also rerun multiple commands
+with ``--since``  and ``--onto`` to specify where to start rerunning from and
+up to which point.
```
kyleam: This sounds off.  I'll describe it in git terms and let you translate as you think is appropriate for this page.

A call `rerun --since=A --onto B REV` reruns the run commits in the range `A..REV` with `HEAD` set to `B` (i.e., `B` is the new base).  It is very much like `git rebase`s `--onto`.

As an added complication, empty strings for --since and --onto are special cases and result in replaying all the run commits from the current branch in another branch.

PR #149, opened 2019-09-04
WIP: Contribution acknowledgements, fixes #140

### LINKED ISSUE #140: Contributor credit?
At the moment the book has a single set of authors. For 3rd-party contributions it might be important to document authorship in a more detailed fashion. Maybe a statement underneath the chapter title would work?

That being said, I would very much prefer to have the whole book being a single editable, evolving entity and not a collection of contributed chapters, where we would have to contact original authors to edit their contributions.

ATM we link the 'contributor list' on the frontpage to https://github.com/datalad-handbook/book/graphs/contributors While this is technically precise, it focuses on the commit of a contribution. But in reality such a contribution could have been the result of debate and evaluation of multiple people.

A dedicated hand-written 'contributors' section could be a solution to both issues: declare original authorship when desired, and acknowledge non-commit contributions.

I have no good idea on how to structure/order such a page. I would like to avoid having to "weigh" contributions in order to determine a subjective importance. The only other two orders I can think of are alphabetical and 'order of appearance' -- and I Iike the latter more.

What do you think?

adswa:
This includes information and contributors, and how to add contributors to our acknowledgments.

### DIFF
--- a/.zenodo.json
+++ b/.zenodo.json
@@ -1,0 +1,38 @@
+{
+  "creators": [
+    {
+      "name": "Hanke, Michael",
+      "affiliation": "Institute of Neuroscience and Medicine, Brain & Behaviour (INM-7), Research Centre Jülich, Jülich, Germany and Institute of Systems Neuroscience, Medical Faculty, Heinrich Heine University Düsseldorf, Düsseldorf, Germany",
+      "orcid": "0000-0001-6398-6370"
+    },
+    {
+      "affiliation": "Institute of Neuroscience and Medicine, Brain & Behaviour (INM-7), Research Centre Jülich, Jülich, Germany",
+      "name": "Wagner, Adina S.",
+      "orcid": "0000-0003-2917-3450"
+    },
+    {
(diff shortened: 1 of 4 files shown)


mih:
Is the allcontributors spec open to arbitrary fields? If somehow possible, we could then use it as canonical source and add affiliation and orcid in there too.

adswa:
It looks like one can define templates in the .all-contributorsrc file (https://allcontributors.org/docs/en/bot/configuration). I'll find out!

adswa:
If I understand it correctly, custom types of contributions are possible. But I have asked [here](https://github.com/all-contributors/all-contributors/issues/301) whether any form of "extended" contribution is possible as well. Let's leave this PR open until we know more.

adswa:
Unfortunately, the template cannot take arbitrary keys, and it is out of the scope of the project, according to the maintainers.

mih:
Hmm, I thought this was merged already. Any reason to keep it WIP? We can deal with a script to automate zenodo.json updates later.

adswa:
Oh, no - also forgot about this. Feel free to merge.

Issue #143, opened 2019-09-04
Student supervision workflow OR archive data/results/...


mih:
This isn't necessarily (only) a standalone use case, but more of an important aspect regarding the evolution/lifetime of a dataset. Anyways, here is a sketch of a workflow that assists in technical aspects of supervising research projects with computational components.

### scenario
- a person is tasked with performing some kind of computational analysis
- the person is a trainee, hence learning on the job
- appropriate supervision is critical to: a) ensure validity of results, and b) make the project a productive learning experience for the trainee

### issues
- time that can be dedicated to supervision is limited. If there are technical complexities/issues they will distract from the "mission". If the "mission" is really important or there is a looming deadline, technical shortcuts tend to be made that turn proper structure and bookkeeping in a project into a mess
- if things are a mess: a) the training result will be suboptimal, b) archiving results is needlessly wasteful, c) re-use is impaired or impossible

### datalad approach
- apply YODA principles
- a (student) project is represented by a dataset
- any "inputs" to a project are tracked as subdataset within that dataset
- any custom code written for the project are version controlled within the dataset
- any third-party software tools are tracked in the dataset as container images
- the student received a properly preconfigured dataset from the supervisor, such that the visible technical complexity is minimized

### benefits
- appropriate data management is key element of the project, not an afterthought that needs to be addressed at the end of its lifetime
- the dataset becomes the lab notebook, hence a valid and detailed log is always available and accessible to supervisor and supervisee
- supervisors can efficiently prepare for meetings in a way that does not rely exclusively on a students report. This shifts the focus from trust in a student to trust in a student's work.
- supervisors can provide feedback, not only high-level based on a presentation, but much more detailed, and also on process aspects
- if desired/necessary: supervisors can directly contribute in a way that is as auditable/accountable as the student's own contributions -- for both parties
- the strict separation and tracking of any external inputs of a project make it possible (when a project is completed) that a supervisor can efficiently test the integrity of the inputs, discard them (if unmodified), and only archive the outputs that are unique to the project -- which then can become a modular component for re-use in a future project

The is obviously a large overlap between this and the 'executable paper' use case. However, the above scenario is valid, even if aspects of the involved work are not computationally reproducible -- regardless of whether that is by design (manual labor), or due to lack of relevant skills. If written up as a use case chapter, it would not need any actual code examples IMHO -- just pointers to the datalad features needed, and a high-level presentation on how the combination of them helps to address the issues.

adswa:
This was added in #153.

PR #152, opened 2019-09-04
[WIP] Supervision use case as proposed in #143

adswa:
Initial draft for a supervision workflow with DataLad. Will cut some superfluous storyline the next time I read through it.


adswa:
uhh, now here I have created an interesting mess. I will close this and cherry-pick what I need on a different branch.

PR #153, opened 2019-09-04
[WIP] Supervision use case as proposed in #143

adswa:
(accidentally rebased #152 into something completely obscure and lazily doing it right on a new branch).

This adds a draft to what @mih proposed in #143. I will cut unnecessary storyline the next time I read it.


mih:
This is precisely what I had in mind. Thx for writing it up! I will make a few comments re possible minor changes later on, when I am at my laptop again.

PR #154, opened 2019-09-04
Change101-103-modify.rst

marisaheckner:
I find this info a little confusing at this step (as it comes two steps after and it gives you the impression that you should already do it now). Therefore I would just delete it.

### DIFF
--- a/docs/basics/101-103-modify.rst
+++ b/docs/basics/101-103-modify.rst
@@ -57,9 +57,6 @@ to accomplish this, but you can take any editor of your choice.
    :workdir: dl-101/DataLad-101
 
    $ cat << EOT >> notes.txt
-   The command "datalad save [-m] PATH" saves the file
-   (modifications) to history. Note to self:
-   Always use informative, concise commit messages.
 
    EOT
 


adswa:
I had to chew on this a bit, because I wonder how to make the intention of this more clear in the book. 

What is supposed to be conveyed in this chapter is how to modify small files and save these modifications. We're embedding this in the narrative by "note-taking", but in reality, it for example would be a workflow when scripting an analysis ('write function' --> save --> 'write another function' --> save --> ...).
So in order to have everyone execute small file modifications, I'm including as many of these "notes" about commands as I can (also to give a concise summary to make everything a bit more memorable). In all other cases, the notes come directly when a new command is explained, but we introduced ``save`` before we were talking about modifications, so there little room to include it earlier. 

Note also the general command structure (I might need to explain what is happening in here in a hidden section -- or if you want to, you could, and submit a PR)

```
   $ cat << EOT >> notes.txt     
```
This line is playing a lot with Unix "[redirection](https://en.wikipedia.org/wiki/Redirection_(computing))" and [here documents](https://en.wikipedia.org/wiki/Here_document). I assume this is very confusing to look at. What happens here is this:
- Take a stream of input in the form of a here document (not yet specified) and send it to the ``cat`` command. ``cat`` just reads what it is given and returns it to the terminal (to be precise: to ``stdout`` -- more on what this is [here](https://en.wikipedia.org/wiki/Standard_streams))
- Redirect everything that comes from ``cat`` into a file ``notes.txt`` by *appending* it (this is what ``>>`` does)
- The EOT is a delimiting character for a stream of input (we could also give it any other name, but EOT I think is a standard and short for "end of text". The here document has the syntax 
``` << EOT something EOT``` and everything between the delimiting characters is passed as an input stream.

```
   The command "datalad save [-m] PATH" saves the file
   (modifications) to history. Note to self:
   Always use informative, concise commit messages.
```
- this is a note as one would take in a course, and because it is between the ``EOT``s it will be passed to ``cat`` (in the formatting as it is here)
```
   EOT
```
- and this terminates the input stream

Do you think it would be more clear on what the intention is if I (or you, or we together) explain how these notes work in one of those foldable sections?

marisaheckner:
Yes I did understand it. Maybe just leave it for now and we can brainstorm at the retreat?
What I was also referring to was the *datalad save -m* that was brought up in this script snippet but also in the next one. So I would just leave it out of the first snippet (where I deleted it) so it is not repeated and thus confusing.
As I see in the handbook you already deleted it but now is says instead: "One can create a new dataset with 'datalad create [--description] PATH'.
The dataset is created empty". I think that is a mistake?

adswa:
Sure, let's leave this open for now and take a look together. I'm not sure I get what is meant by this


> As I see in the handbook you already deleted it but now is says instead: "One can create a new dataset with 'datalad create [--description] PATH'.
The dataset is created empty". I think that is a mistake?

Maybe we use the luxury of in-person interaction at the retreat for explanations and brainstorming :)

marisaheckner:
Yes lets do that! :-)

adswa:
Closing this after clarification with @marisaheckner

PR #156, opened 2019-09-04
Update101-103-modify.rst

marisaheckner:
As mentioned in my earlier comment: I would mention this when you explain git log for the first time and then delete it here.

### DIFF
--- a/docs/basics/101-103-modify.rst
+++ b/docs/basics/101-103-modify.rst
@@ -82,7 +82,7 @@ and save the file:
 Let's take another look into our history to see the development of this file.
 We're using :command:`git log -p` to see the difference to the previous state of a
 file within each commit. (Note: the output below is an excerpt, if you enter
-the git log, your history will be longer. 
+the git log, your history will be longer, and thus opened in a [pager](https://en.wikipedia.org/wiki/Terminal_pager). You can get out of it by pressing ``q``.
 
 .. runrecord:: _examples/DL-101-103-107
    :language: console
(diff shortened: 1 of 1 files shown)


docs/basics/101-103-modify.rst
```diff
@@ -82,8 +82,7 @@ and save the file:
 Let's take another look into our history to see the development of this file.
 We're using :command:`git log -p` to see the difference to the previous state of a
 file within each commit. (Note: the output below is an excerpt, if you enter
-the git log, your history will be longer. You can get out of ``git log`` by pressing
-``q``.)
+the git log, your history will be longer.
```
adswa:
```suggestion
the git log, your history will be longer, and thus opened in a [pager](https://en.wikipedia.org/wiki/Terminal_pager). You can get out of it by pressing ``q``.
```

adswa:
Thanks a lot for engaging with the book and submitting PRs!
Following up on my comment in #154, I'd like to keep the information on how to get out of the pager in this chapter, because in most cases, this is the first time people will end up in it, and it is good to have the information nearby. However, it might be helpful to mention in more detail what is happening and why that might differ from the previous git log (that likely was without a pager). If you're fine with it, I'd suggest the change proposed in the review, even if it is not what you originally intended with the PR (as before, you can commit it right in this interface, or add it and push it yourself, and I'd be happy to show you next week). Let me know what you think.

adswa:
Please decide whether you want to have this with suggested change as well or have it stay closed. Just as in the other PR, you can reopen and commit, if you want

adswa:
Thx! (I noticed I made a mistake in my suggestion... not enough coffee, I suppose. I'll add a commit on top of this, and then I'll merge)

Thanks again for the PRs! :)

PR #165, opened 2019-09-11
Modernize versioning setup.

mih:
- inspired by https://gist.github.com/effigies/c9f4194034ee218bab1668bfd7851cfc
- use versioneer `__version__` of support package for the book itself

Now all that is needed to have a new/sane version is a git tag


adswa:
thats fancy! thx!

PR #166, opened 2019-09-11
Hand-polish contributor name

mih:
Q is now, if the bot can handle to when the next update happens.


adswa:
That will be interesting. Thx for the fix.

PR #168, opened 2019-09-11
Revert even more setup.cfg setup to get readthedocs to install deps

mih:


mih:
Started working again.

PR #172, opened 2019-09-13
Improve sidebar experience, fixes #170

### LINKED ISSUE #170: Sidebar taller than screen, doesn't scroll
In Firefox 68.0.2 on OSX, the sidebar has a large chunk of white space, pushing some of the contents off screen. Looks like #62 set the sidebars to fixed, which means those contents can't be scrolled to.

<div display="flex">
Screen Shot 2019-09-12 at 11 34 07
Screen Shot 2019-09-12 at 11 34 13
</div>

adswa:
As noted by @effigies in #170 (thanks!), content in the sidebar can be invisible, for example when zooming in too much. This is both because a larg-ish empty box makes it taller than the page, and because the side bar is fixed.

In principle, I like the space in the sidebar because it creates a nice visual offset between the intro in the sidebar and the Contributors section, making both parts more striking/visible. I'd also like to have the sidebar fixed in the sense that scrolling the main content does not scroll the sidebar away, because it keeps the navigation links to the landing page, pre-/succeeding sections, Github, and the quicksearch close by in very long sections. 
In this PR I therefore try to mitigate the problem by

- decreasing the amount of white space between intro and contributors in the sidebar
- giving the side bar its own vertical scroll bar as soon as it does not fit on a single page anymore

@mih, do you have thoughts on this?

### DIFF
--- a/docs/_static/custom.css
+++ b/docs/_static/custom.css
@@ -42,3 +42,8 @@ div.gitusernote {
     background-color: #c3ebffff;
     border: 1px solid #7fd4ffff;
 }
+
+div.sphinxsidebar {
+    max-height: 100%;
+    overflow-y: auto;
+}
--- a/docs/_templates/sidebarintro.html
+++ b/docs/_templates/sidebarintro.html
@@ -26,7 +26,7 @@
 </p>
 
 <div style="display:block;position:relative;margin: 1em 0 1em 0;">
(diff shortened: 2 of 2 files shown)


mih:
I think this is a good and practical solution. At some point in the future, we will want to take a more global look at the site and will probably make some changes. Until then...

effigies:
Just a note that I finally checked and this works fine for me. Thanks!

PR #176, opened 2019-09-14
Links, glossary, index fixes

adswa:
This mostly fixes not-working or missing links/refs, and adds some missing glossary and index entries.


mih:
Thx!

Issue #177, opened 2019-09-14
Non-root (e.g. HPC) installation exploration


mih:
This seems to be trivial with miniconda (thx @effigies for the hint, and @yarikoptic  for putting the pieces into conda). Here is the recipe:

```sh
$ wget https://repo.anaconda.com/miniconda/Miniconda3-latest-Linux-x86_64.sh
$ bash Miniconda3-latest-Linux-x86_64.sh
# acknowledge license, keep everything at default
$ conda install -c conda-forge datalad

# result
$ datalad --version
datalad 0.11.7
$ git-annex version
git-annex version: 7.20190819-ge4cecf8
...
$ git --version
git version 2.20.1

# fully populated conda installation
$ du -sh miniconda3/
856M    miniconda3/
```

The installer automatically configures the shell to make conda-installed tools accessible, so no further configuration is necessary. Such a snippet is added to `.bashrc` (all paths, here and below, have been redacted):

```
...
# >>> conda initialize >>>
# !! Contents within this block are managed by 'conda init' !!
__conda_setup="$('/HOME/miniconda3/bin/conda' 'shell.bash' 'hook' 2> /dev/null)"
if [ $? -eq 0 ]; then
    eval "$__conda_setup"
else
    if [ -f "/HOME/miniconda3/etc/profile.d/conda.sh" ]; then
        . "/HOME/miniconda3/etc/profile.d/conda.sh"
    else
        export PATH="/HOME/miniconda3/bin:$PATH"
    fi
fi
unset __conda_setup
# <<< conda initialize <<<
```

The installed datalad version is not suitable for the handbook yet, but 0.12 is coming.

@dickscheid this approach should avoid tedious hand-deployment, give immediate accessibility with requiring per-dataset reconfiguration, and also offers built-in update-ability. Proper functioning verified on a JURECA login node (minus two failing tests, see below).

```
$ datalad test      
[INFO   ] Starting test run for module(s): ['datalad'] 
Running unit tests for datalad
Python version 3.7.3 (default, Mar 27 2019, 22:11:17) [GCC 7.3.0]
nose version 1.3.7
...
======================================================================
ERROR: datalad.support.tests.test_annexrepo.test_files_split_exc(<class 'datalad.support.annexrepo.AnnexRepo'>,)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/HOME/miniconda3/lib/python3.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
[...]

PR #178, opened 2019-09-16
add nonroot linux installation

adswa:
Based on the description in #177 , this adds a short, toggled container with instructions for non-root access Linux machine. We can keep it open for potential adjustments if more comes out of #177.


mih:
I am undecided whether this should be hidden by default. One could argue that non-root access is the standard, whenever we are talking about "not-my-own-machine". However, I don't feel strongly about it, so let's get it in first, and possibly reconsider later.

In any case: Thanks for writing it up!

adswa:
Sounds good! Thx!

PR #194, opened 2019-10-02
Update 101-106-nesting.rst

lisanmo:
correction of typo, "DataLad-101" was referenced as the subdataset while it is the superdataset (if I understood correctly)


adswa:
Hey, @all-contributors, please add @lisanmo for bug and content!

Issue #193, opened 2019-10-02
Add a "progress bar" or keep contents visible


lisanmo:
**Is your feature request related to a problem? Please describe.**

When I was working my way through the "Basics" part (which I loved - I could follow along easily and even felt well entertained!) I got to the point where I had to decide to go home or keep working - I would have loved to see directly on the page I was where I was within the "Basics" section. Would I need only 10 minutes to finish it or a few more hours? Would this be a good point to take a break?
At the side and the bottom of the page I could only see the previous and next chapters, but no overview.

**Describe the solution you'd like**
I would like to have a kind of progress bar or overview of where exactly I am in the section or even entire book, so I can see more of what will come next and plan ahead.

**Describe alternatives you've considered**

I could have just opened the "Documentation overview" on the left of the page and return to the table of contents, find which chapter I am working on there and seen what is to come.

However,
a) that feels more disruptive to me than being able to check it while I remain on the site I am working on. Maybe the "Related Topics" section could be designed to expandable? Then you can minimise it if you don't want to be distracted by a full list of contents, but get an easy overview whenever you feel like it.
b) when following the book, I felt like I was in the "Basics" section and expected an "Advanced" section to follow upon completion. When opening the "Documentation overview", the section I was working on was now called "DataLad datasets". Because the chapters are listed there, it is no problem for navigation, but looking back it feels a bit disjunct (language-wise) from the "Basics" section the handbook leads you to. 

**Additional context**

I hope I could make clear what I mean! If not, please let me know!

adswa:
Hey @lisanmo, thanks a lot for the feedback and the solutions you propose!
I have tried to play with both (full table of contents, progress bar). Displaying a full table of contents in the sidebar didn't work so well:

![image](https://user-images.githubusercontent.com/29738718/66068108-41c23580-e54d-11e9-9b74-62be2f56a340.png)

I'm inclined towards the progress bar, and I've been playing with it briefly (see  #196). So far in my toy examples I only have a progress bar per page (and I believe you'd rather have a progress bar related to a full "chapter" (e.g. all of DataLad's datasets - Exploring DataLad's core data structure, if I understand correctly):

![image](https://user-images.githubusercontent.com/29738718/66068385-c745e580-e54d-11e9-895c-68fa266d9f6e.png)

But I'm happy to extend my css/html/javascript a bit with this and find out whether I can make chapter-wise progress bars.

[...]

PR #196, opened 2019-10-02
Page-wise progress bars

adswa:
After #193 I looked a bit into "scrolling progress bars". This isn't yet what @lisanmo suggests (it's only page-wise) and I'm not sure yet whether or how one could get a chapter-wise progress bar in the HTML version of the book, but I will continue exploring. This is how it currently looks like (The blue bar progresses the more you scroll):

image

### DIFF
--- a/docs/_templates/sidebarlogo.html
+++ b/docs/_templates/sidebarlogo.html
@@ -20,3 +20,5 @@
   height: 100%;
 }
 </style>
+
+<progress id="progressbar" value="0" max="100"></progress>
--- a/docs/_templates/page.html
+++ b/docs/_templates/page.html
@@ -12,3 +12,18 @@
     });
 </script>
 {% endblock %}
+
+{% block header %}
+  <script type="text/javascript">
+
+  $(window).scroll(function () {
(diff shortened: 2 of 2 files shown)


docs/_templates/sidebarlogo.html
```diff
@@ -20,3 +20,5 @@
   height: 100%;
 }
 </style>
+
+<progress id="progressbar" value="0" max="100"></progress>
```
adswa: brief thought to follow up later: Maybe its possible to add this progress bar page-wise into the header, and set ``value`` and ``max`` based on a counter in each chapter. A chapter with 5 sections would add one fifth to the progress bar per page read (while pace of progress bar for that fifth is still dependent on section length).

adswa:
Having played with this a bit more, I failed to implement a progress bar that spans chapters. It may well be possible, but for now I'm giving up. As it currently is, this PR would add a page-wise progress bar. IMO it's a nice-to-have but not a need-to-have.

mih:
+1

adswa:
(if nobody complains within the next day or so I will let us have page-wise progress bars)

Issue #189, opened 2019-09-28
shorten full paths in the book?


yarikoptic:
I guess we are to deal with those full paths at datalad level eventually (e.g. reduce use of full paths in the logs and Dataset's `repr` outputs), but I wondered if may be the book examples could be built/ran under some other shorter path than e.g. `/home/adina/repos/datalad-handbook/docs/_build/`, may be `/tmp/book` or alike?
That would also assure that book build/examples ran by different ppl would produce the same result/do not introduce a diff

adswa:
I agree that the long paths can be confusing (and also make code snippets harder to read). I think  https://github.com/datalad-handbook/book/issues/53 is related/similar. I'm not an expert on this so I don't know whether it is possible/useful to have the workdir set up outside of the handbooks directory (@mih wrote the ``autorunrecord`` extension module we are using to set up working directories to execute code in). Currently, all code that needs a working directory is under ``/datalad-handbook/docs/_build/wdirs``, which assures that anyone can execute/build the book (relative path within the ``handbook`` directory), and also that the datasets and their content in the workdir lasts (as opposed to anything in ``/tmp/``; rerunning everything from scratch already takes a few minutes, it is immensely helpful to not have to rerun everything all of the time).

I wonder whether it may just be sufficient to have ``sed`` replace all ``/home/adina/repos/datalad-handbook/docs/_build/wdirs/`` in the examples with ``/home/someuser`` (or something similar) in the Makefile for a "quick fix", @mih?

mih:
I'll look into it.

Issue #53, opened 2019-07-05
Uniform user environment for the book examples


mih:
At the moment the current system user environment (i.e. us) is used. That means that when I re-run and replace @adswa 's output, there will always be changes.

We could include a helper script that create a dedicated user enironment on our machines that we then use to build the book examples.

adswa:
It sounds good to not replace output when re-running ``make`` after a ``make clean``!

I have to admit, though, that I don't know exactly what the dedicated user environment would do? Create a "common author" so that paths and commit authors always are the same?

mih:
Yes, exactly. It would essentially be a (fake) HOME directory with a uniform identify, Git configuration, shell prompt setup. We have a similar setup for build the asciicasts in DataLad. I can work on that at some point.

Side note: the output would still be replaced, but changes would hopefully only reflect difference in software versions, and not in user identity or configuration.

adswa:
Cool!

PR #198, opened 2019-10-05
Run code examples in a sanitized environment

mih:
- [x] needs autorunrecord 0.4 or later (requirements adjusted)
- [x] rebuild recorded examples (works for me, but better if done
      by someone else too for verification (@adswa)
- [x] adjust `clean-examples` target in `Makefile`, but I am reluctant to say `rm -rf /home/me/*`
      Should we have another directory inside `/home/me`? But it would make all paths longer. However, this is a small price for a future contributor with a `me` useraccount...
- [x] fixes gh-53
- [x] fixes gh-189

In order to not have to run everything in a tmp directory, I decided to
have a proper alternative HOME dir -- this will look more familiar.

This requires manual setup (on each machine that shall build code
examples, i.e. not on readthedocs):

```
sudo mkdir /home/me
sudo chown $USER:$USER /home/me
```

until https://github.com/datalad/datalad/issues/3750 is fixed, it also
requires:

```
HOME=/home/me git config --global user.name "Elena Piscopia"
HOME=/home/me git config --global user.email "elena@example.net"
```

Example diff

```diff
--- a/docs/basics/_examples/DL-101-101-101
+++ b/docs/basics/_examples/DL-101-101-101
@@ -1,6 +1,6 @@
 $ datalad create --description "course on DataLad-101 on my private Laptop" -c text2git DataLad-101
-[INFO] Creating a new annex repo at /home/adina/repos/datalad-handbook/docs/_build/wdirs/dl-101/DataLad-101 
+[INFO] Creating a new annex repo at /home/me/dl-101/DataLad-101 
 [INFO] Running procedure cfg_text2git 
 [INFO] == Command start (output follows) ===== 
 [INFO] == Command exit (modification check follows) ===== 
-create(ok): /home/adina/repos/datalad-handbook/docs/_build/wdirs/dl-101/DataLad-101 (dataset)
+create(ok): /home/me/dl-101/DataLad-101 (dataset)
```


adswa:
This works for me as well, I've ticked the verification TODO. Thanks much for doing this!

I share your reluctance about ``rm -rf me`` (it seems like a not-completely-implausible username). What about something like ``/home/me/dl``? 

I'll also
- [ ] update ``contributing.rst`` accordingly

mih:
Great! Thanks for trying it out!

> What about something like /home/me/dl

That would not be enough, but maybe:

```
% git grep ':workdir:' | cut -d ':' -f 4- | sort | uniq | cut -d '/' -f 1 | uniq
 dl-101
 studyforrest
 usecases
```

adswa:
> That would not be enough, but maybe:
> 
> ```
> % git grep ':workdir:' | cut -d ':' -f 4- | sort | uniq | cut -d '/' -f 1 | uniq
>  dl-101
>  studyforrest
>  usecases
> ```

I'm not sure I understand what you mean. Do you mean removing those three subdirectories individually in the Makefile (instead of ``/home/me/*``)? If we set ``/home/me/dl`` as what was formerly ``docs/_build/wdirs`` or now the ``autorunrecord_basedir``, during ``make`` those three directories would be created, and ``rm -rf /home/me/dl/*`` would get rid of them -- or do you mean ``/home/me/dl`` is still possibly an existing directory?

mih:
Ah, sorry, misunderstood. My only point was to limit the `rm -rf` to the actual directories that would be created by the examples we have at any point. That minimizes the risk, and is orthogonal to any additional constraints we might want to put on the naming of the workdirs.

Apart from the fact that deleting more directories means more risk, I don't think deleting `dl` is particularly safer than deleting `dl-101`.

adswa:
Ah, got it. Sure, sounds sensible!

mih:
Thx for the write up!

mih:
I have updated the Makefile and am trying to recompute all examples.

mih:
Updates are pushed.
There is one open issue: We are using `ls -l` in some places that
reveals that the effective user ID creating files on the machine is
not the fake one we configured. We should check whether it is important
to list this information. If so, we may need to configure a dedicated
user.

mih:
Good to go from my POV. @adswa please merge when you are happy with it.

adswa:
FTR: unfortunately, this does not yet work on brainbfast (i.e. machine I do not have root access to). My current workaround is to downgrade to ``autorunrecord==0.3``. 

Unrelated FTR: Book examples work fine with ``0.12.0rc5``

PR #206, opened 2019-10-08
adjust latex fonts to resemble those of readthedocs

adswa:
This lets the PDF use fonts that are closer to what readthedoc's theme uses. It's pure personal taste - I find those fonts more readable and more modern, but I'm also happy to stick to the defaults.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ python:
 cache:
   - apt
 install:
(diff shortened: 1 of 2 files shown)


adswa:
This required adding ``texlive-fonts-recommended`` to the travis setup, which slows down the continuous integration a bit...

mih:
Cool, thx.

PR #207, opened 2019-10-08
Update 101-115-symlinks.rst

lisanmo:
In line 136 the sentence is left without an ending, but now I'm too curious to hear about the second advantage!

P.S. I could not propose a file change without actually changing the file, so I added a question mark.

### DIFF
--- a/docs/basics/101-115-symlinks.rst
+++ b/docs/basics/101-115-symlinks.rst
@@ -133,7 +133,7 @@ is only one single piece of data in your object tree. This, depending on
 how much identical file content lies in different parts of your dataset,
 can save you much disk space and time.
 
-The second advantage is a
+The second advantage is a ?
 
 .. gitusernote::
 


adswa:
Hehe, this is a bit of a "wordplay": The second advantage is the gitusernote. I.e., its meant to be read as:

"The second advantage is a Note for Git users: Small symlinks can be written very very fast when switching branches, as opposed to copying and deleting huge data files."

I wonder how often people stumble across this and whether it would be better to add

"The second advantage is in the following note for Gitusers:" 

:thinking:

lisanmo:
Oh! Now I see it. Sorry for the false alarm 😄 
Maybe you can wait and see if more people note on it or not, might be I'm just the outlier.

adswa:
That sounds sensible, I will close this for now and we'll revisit once more people stumble. :slightly_smiling_face:

PR #208, opened 2019-10-10
Fix: add missing dl-save

adswa:
Forgot to add a datalad save after modifications of ``notes.txt``.


adswa:
Tests pass, will merge.

Issue #132, opened 2019-08-29
Windows native installation/usage exploration


mih:
FTR: This is what I find to be the simplest installation procedure on Win10 after some trial-and-error

### Install Conda
- pick latest PY3 installer: https://docs.conda.io/en/latest/miniconda.html
- keep everything on default (do not add to PATH)
- this will also install Python
- from now on any further action must take place in the "Anaconda Prompt" (a preconfigured terminal shell)

### Install Git
- `conda install -c conda-forge git` (must be from conda-forge, anaconda version does not provide `cp`)

### Install git-annex
- Download installer from https://downloads.kitenet.net/git-annex/windows/current/
- Install into the miniconda `Library` directory, e.g. C:\Users\mih\Miniconda3\Library

### Install datalad
- it is possible to install datalad via conda too (`conda install -c conda-forge datalad`) but it only supports the 0.11 version, hence is irrelevant in this context
- install datalad via `pip`  as usual

How functional datalad actually is on Win10 if installed this way is subject to further exploration.

### (Confirmed) features

- [x]  OpenSSH:  `OpenSSH_7.9p1, OpenSSL 1.1.1a  20 Nov 2018`
- [x] git-annex: `All 168 tests passed`
- [ ] `datalad.support.tests.test_gitrepo` https://github.com/datalad/datalad/issues/3635
- [ ] `datalad.support.tests.test_annexrepo` https://github.com/datalad/datalad/issues/3636
- [ ] `datalad.core` tests https://github.com/datalad/datalad/issues/3633
- [ ] `datalad.distribution.tests.test_clone` https://github.com/datalad/datalad/pull/3640
- [x] `datalad.distribution.tests.test_install`
- [ ] `datalad.distribution.tests.test_get` https://github.com/datalad/datalad/issues/3634
- [ ] `datalad.support.tests.test_ssh*`

mih:
Installing this way does not provide the `cp` command, which is needed by git-annex. It would be installed by git4win. Conda itself does not seem to provide a `coreutils` package for windows. This needs to be resolved on way or another.

mih:
We might want to consider recommending the installation of https://tortoisegit.org/

It provides Git integration into the explorer (unclear if anyhow useful for us), its startup wizzard support SSH key generation, and identity setup.

mih:
We have now settled on going largely native by default. This should not prevent or discourage people from alternative routines that work for them.

Issue #114, opened 2019-08-19
Windows WSL2 installation/usage exploration


mih:
On a fresh Win10, immediately after install:

### Enable WSL
At the moment, one needs to join the Windows Insider Programm to get access to a build version that has WSL2.

Start the Power Shell as an Administrator.
Run both commands, only restart after the second one (despite being prompted after the first one already)

```
Enable-WindowsOptionalFeature -Online -FeatureName VirtualMachinePlatform
Enable-WindowsOptionalFeature -Online -FeatureName Microsoft-Windows-Subsystem-Linux
```

### Install Debian app

Same procedure as for WSL1

### Enable WSL2 for the Debian app

*Could not continue here, because I tried in a VirtualBox VM, and Win10 in the VM needs proper hardware virtualization for WSL2, which VirtualBox cannot provide (nested) virtualization on Intel CPUs (AMD would be fine).*

##### Next attempt with ``bnbdatalad``:

To set the WSL version to WSL2, run this command as an administrator in the Power Shell:

```
wsl --set-default-version 2
```
In the Microsoft store, search for a Debian distribution. Download and install it, then start it. Pick a user name, and a root password (repeat the password when prompted).

- configure the distro to use WSL 2: Run this command as an administrator in the Power Shell:

```
wsl -l -v
```
This should say something like this (the important detail is ``VERSION 2``):

```
    NAME        STATE               VERSION
*   Debian       Running            2
```

- To enable the Neurodebian repository to install ``datalad`` with ``apt-get``...

    - In your Debian distribution, find out which version you are running -- should be buster (e.g. with``cat /etc/*release`` )
    - install a few missing tools: ``sudo apt-get install wget``, ``sudo apt-get install gnupg`` (or ``gnupg2``)
    - [Enable the Neurodebian repository](http://neuro.debian.net/).
    - ``sudo apt-get update``, ``sudo apt-get upgrade``
    - (one could do ``sudo apt-get install datalad`` now, but would not be ``datalad.0.12``)

- To install ``datalad`` with ``pip``...
    - install python3-pip with ``sudo apt-get install python3-pip``
    - to install DataLad 0.12 from master: ``pip3 install --user git+https://github.com/datalad/datalad.git#egg=datalad`` (but requires a prior ``sudo apt-get install git``)
    - add ``~/.local/bin`` to the ``PATH``
    - to install ``git-annex``: ``sudo apt-get install git-annex``
 
:tada:

PR #211, opened 2019-10-11
finalize installation section

adswa:
This should finally get the installation instructions into a complete-ish state:

- updates Windows native installation instructions with Miniconda (using notes from https://github.com/datalad-handbook/book/issues/132) and make it the default
- add WSL2 instructions from issue for completeness (based on notes in https://github.com/datalad-handbook/book/issues/114)
- removes todos
- add "eg HPC systems" to Linux machines with no root access

Question to @mih: 
- You briefly mentioned enabling SSH Server via powershell yesterday. Am I right in my assumption that this is not necessary for a user-installation (but rather to use ``bnbdatalad``, see https://github.com/psychoinformatics-de/org/issues/91)?
- You once mentioned https://tortoisegit.org/ as an alternative, what do you think about this atm?

### DIFF
--- a/docs/intro/installation.rst
+++ b/docs/intro/installation.rst
@@ -169,7 +169,7 @@ please `get in touch <https://github.com/datalad-handbook/book/issues/new>`_.
 
       - During installation, you will be prompted to "Choose Install Location".
         **Install it into the miniconda Library directory**, e.g.
-        ``C:\Users\mih\Miniconda3\Library``.
+        ``C:\Users\me\Miniconda3\Library``.
 
    - **Step 4**: Install DataLad via pip
 
--- a/docs/intro/installation.rst
+++ b/docs/intro/installation.rst
@@ -44,7 +44,7 @@ DataLad and all of its software dependencies (including the Git-annex-standalone
 
    .. container:: header
 
-      **Linux-machines with no root access**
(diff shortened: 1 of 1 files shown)


mih:
> You briefly mentioned enabling SSH Server via powershell yesterday. Am I right in my assumption that this is not necessary for a user-installation

Yes, correct, this is just for a box that needs to run the tests.

> You once mentioned https://tortoisegit.org/ as an alternative, what do you think about this atm?

I have only explored it briefly. I would not make any point in favor or against it ATM.

docs/intro/installation.rst
```diff
@@ -103,60 +103,97 @@ a user's home directory:
...
+   - **Step 3**: Install Git-annex
+
+      - Obtain the current Git-annex versions installer
+        `from here <https://downloads.kitenet.net/git-annex/windows/current/>`_.
+        Save the file, and double click the downloaded
+        :command:`git-annex-installer.exe` in your Downloads.
+
+      - During installation, you will be prompted to "Choose Install Location".
+        **Install it into the miniconda Library directory**, e.g.
+        ``C:\Users\mih\Miniconda3\Library``.
```
mih: s/mih/me for consistency?

adswa: good catch!

PR #214, opened 2019-10-18
On DataLad versions in installation instructions

adswa:
This warns about 0.12 being not yet available via most package managers, and guides users to install the most recent release candidate with pip- or conda-based methods.

I have also placed the instructions for non-root access Linux machines outside of a toggle-able container.


mih:
Thx

PR #216, opened 2019-10-22
warn about invisible broken symlinks in certain viewers

adswa:
@loj and @bpoldrack discovered that OSX Finder was not able to
display broken symlinks, essentially hiding available files.
This should make readers aware in the section installing datasets,
and give some more information on this in the section on symlinks

### DIFF
--- a/docs/basics/101-105-install.rst
+++ b/docs/basics/101-105-install.rst
@@ -305,6 +305,17 @@ modification.
    $ datalad save -m "Add note on datalad install"
 
 
+.. note::
+
+  Listing files directly after an installation of a dataset will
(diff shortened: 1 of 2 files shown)


mih:
Maybe add a more positive spin on it and suggest upgrading the finder: https://github.com/andrewringler/git-annex-turtle

Sidenote: `git annex adjust` also has the ability to explicitely hide missing files, so in some cases this can be desired behavior.

adswa:
Thanks for the pointer, I included it and tried to give it a more positive spin - @loj, maybe this extension will be useful for Susanne et al.

adswa:
Thx!

PR #219, opened 2019-10-23
Proof-of-concept tests for autorunrecord2casts

adswa:
Includes currently relevant changes to `conf.py` and some tags. I assume builds will fail due to un-updated autorunrecord atm.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,11 +6,10 @@ build: html
 %:
 	cd docs && make $@
 
-# wipe out all recorded examples and casts
+# wipe out all recorded examples
 clean-examples:
 	# check if we have something like .xsession or a .bashrc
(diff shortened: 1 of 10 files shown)


adswa:
This pull request is now
- updated requirements.txt to precise commit in `autorunrecord` that introduced the casting capabilities
- turns the first chapter (on datasets) into a cast dataset_basics. The call requires a ``CAST_DIR`` env variable to be set to any path where casts should be stored
- has all relevant annotations in the first chapter (might fine tune them in preparation for course, but works for now)
- tiny fixes to make cast script run

From my side its good to merge. More like this will come as annotations for further casts.

mih:
Cool, thx!

PR #226, opened 2019-10-29
Python API

adswa:
Together with elaborating briefly on DataLad's Python API, this section now contains a minimal YODA analysis.

Leftover TODOs:
- [x] adjust file and example numbering
- [x] resolve TODOs in text
- [x] add a section (?) on "advanced dataset nesting" that showcases how changes in a subdataset look in the superdataset. This *needs* to happen prior to merging this, so I will add it into this PR.
- [x] add a section on sharing an analysis like this (my idea: via Github). 
- [x] language tweaks

#242 is a candidate for extending the analysis performed in here with including the software environment and using ``datalad containers-run``.


docs/basics/101-130-yodaproject.rst
```diff
@@ -0,0 +1,564 @@
...
+DataLad dataset.
+
+For your submission, you decide to analyse the
+`iris flower data set <https://en.wikipedia.org/wiki/Iris_flower_data_set>`_.
+It is a multivariate data set on 50 samples of each of three species of Iris
+flowers (*Setosa*, *Versicolor*, or *Virginica*), with four variables: the length and width of the sepals and petals
+of the flowers in centimeters. It is often used in introductory data science
+courses for statistical classification techniques in machine learning, and
+widely available, among many other sources as
+`Github gists <https://gist.github.com/netj/8836201>`_.
```
mih: This is a good source that also comes with code and a plot: https://scikit-learn.org/stable/auto_examples/datasets/plot_iris_dataset.html

mih: Ah, I understand now that the file is the important part...

docs/basics/101-130-yodaproject.rst
```diff
@@ -0,0 +1,564 @@
...
+After the last lecture, you naturally want your dataset to follow the YODA
+principles. As a start, you use the ``cfg_yoda`` procedure to help you
+structure the dataset [#f1]_.
+
+.. runrecord:: _examples/DL-101-130-105
+   :language: console
+   :workdir: dl-101/DataLad-101
+
+   $ cd midterm_project
+   $ datalad run-procedure cfg_yoda
```
mih: I would make sense to me to to this simultaneously with `create`, but it may be just my taste.

docs/basics/101-130-yodaproject.rst
```diff
@@ -0,0 +1,564 @@
...
+   from sklearn import model_selection
+   from sklearn.neighbors import KNeighborsClassifier
+   from sklearn.metrics import classification_report
+   from datalad.api import get, install
+
+   data = "input/iris.csv"
+
+   # make sure that the data is obtained:
+   install('input/')
+   get(data)
```
mih: When the subdataset is properly linked (like it is here), there is no need for `install()`. The `get(data)` alone will do!

docs/basics/101-130-yodaproject.rst
```diff
@@ -0,0 +1,564 @@
...
+
+   # prepare the data as a pandas dataframe
+   df = pd.read_csv(data)
+   attributes = ["sepal_length", "sepal_width", "petal_length","petal_width", "class"]
+   df.columns = attributes
+
+   # create a pairplot to plot pairwise relationships in the dataset
+   plot = sns.pairplot(df, hue='class')
+   plot.savefig('output/pairwise_relationships.png')
```
mih: I would prefer to use something like `figures/` instead of output, to emphasize that the entire top-level dataset is for the output.

docs/basics/101-130-yodaproject.rst
```diff
@@ -0,0 +1,564 @@
...
+
+.. runrecord:: _examples/DL-101-130-113
+   :language: console
+   :workdir: dl-101/DataLad-101/midterm_project
+
+   $ datalad run -m "analyze iris data with classification analysis" \
+     --input "input/iris.csv" \
+     --output "output/prediction_report.csv" \
+     --output "output/pairwise_relationships.png" \
+     "python3 code/script.py"
```
mih: In principle `code/script.py` is also "input". Only because we know that it is tracked in Git within this very dataset, we can skip that. If we listed it as input, DataLad would also obtain potentially required subdatasets with code (or entire toolboxes) on demand.

docs/basics/101-130-yodaproject.rst
```diff
@@ -0,0 +1,564 @@
...
+.. index:: datalad command; save --to-git
+
+One feature of the YODA procedure was that it configured certain files (for
+examples everything inside of ``code/`` and the ``README.md`` file in the
+root of the dataset) to be saved in Git instead of Git-annex. This was the
+reason why the ``README.md`` in the root of the dataset was easily modifyable.
+However, such a configuration does not exist for the ``README.md`` file we
+are creating in ``output``. This means, should we :command:`datalad save`
+this file, it will be annexed -- an inconvenience, given that this file is
+small enough to be handled by Git.
```
mih: Yeah, this is what I mean. It is better to consider the entire dataset as output, and then the one top-level README can contain everything there is to know. Just my 2ct

docs/basics/101-130-yodaproject.rst
```diff
@@ -0,0 +1,564 @@
...
+
+   The upcoming part requires a Github account. If you do not have one you
+   can either
+
+   - Create on now -- it is fast, free, and you can get rid of it afterwards,
+     if you want to.
+   - Or exchange the command ``create-sibling-github`` with
+     ``create-sibling-gitlab`` if you have a Gitlab account instead of a Github
+     account.
+   - Don't listen to me and not follow along. I'm only a book, not your mom.
```
mih: ;-)

docs/basics/101-130-yodaproject.rst
```diff
@@ -0,0 +1,564 @@
...
+
+      $ cd ../../
+      $ datalad install -r --source "https://github.com/adswa/midtermproject.git"
+
+   Note that we performed a *recursive* installation by providing the ``-r``
+   option. Thus, we don't need to install the ``input/`` subdataset anymore.
+   Let's start with the subdataset, and see whether we can retrieve the
+   input ``iris.csv`` file. This should not be a problem, since it's origin
+   is recorded:
+
```
mih: Unclear why this works. I think it might be only because we created the iris dataset manually above. It might be more useful if we put that dataset on github directly, and install it from there to begin with?

docs/basics/101-130-yodaproject.rst
```diff
@@ -0,0 +1,564 @@
...
+   .. runrecord:: _examples/DL-101-130-128
+      :language: console
+      :workdir: dl-101/midtermproject
+      :realcommand: echo "datalad rerun $(git rev-parse HEAD~2)" && datalad rerun $(git rev-parse HEAD~2)
+
+   Hooray, your analysis was reproduced! Note however that none of the DataLad
+   commands would have been necessary to reproduce the analysis -- simply
+   executing the Python script would have installed and retrieved the inputs thanks
+   to the DataLad functions used in the script, as long as DataLad is installed
+   on the system.
```
mih: This is a bit confusing. DataLad is not needed, but DataLad needs to be installed. I think it would be good to emphasize that the rerunning part did not even require knowledge about the analysis. It might be possible to factor out the `HEAD~2` as well, and just use a tag. It would look even more generic then.

docs/basics/101-131-summary.rst
```diff
@@ -0,0 +1,28 @@
+.. _summary_python:
+
+Summary
+-------
+
+DataLad's Python API makes all of DataLad's functionality available in
+Python, either as standalone functions that are exposed via ``datalad.api``,
+or as methods of the ``Dataset`` class.
+This provides an alternative to the command line, but it also opens up the
+possibility of performing DataLad commands directly inside of scripts -- a
```
mih: The somewhat indicates that only Python can do script. But Matlab users can use datalad as well (via standard system calls) -- just not that convenient.

docs/basics/101-130-yodaproject.rst
adswa: FTR I published the dataset to Github once and then replace the interactive code snippet that was executed with a static code-block. But the dataset currently lives in ``github.com/adswa/midtermproject``.

adswa:
(pushed a first batch of fixes, more to come)

docs/basics/101-129-python.rst
```diff
@@ -0,0 +1,79 @@
...
+
+Use cases for DataLad's Python API
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+"Why should one use the Python API? Can we not do everything necessary via the
+command line already? Does Python add anything to this?" asks somebody.
+
+It is completely up to on you and dependent on your preferred workflow
+whether you decide to use the command line or the Python API of DataLad for
+the majority of tasks. Both are valid ways to accomplish the same results.
+One typical application though is the use of :command:`datalad get` in Python
```
mih: Everything that follows is more an example of what can "also be done in Python". I see the danger that it comes across as "better be done in Python", and consequently communicates that Python is the "favorite child" -- which it really isn't.

There is one major advantage of using the Python API though, and it is exactly the one you start with: the Dataset. Via the command line a dataset needs to be discovered and inspected with each datalad command call, because there is no persistence of information in a, e.g., shell script. Given that the `datalad` command has a startup time (even when doing nothing) of ~200ms, this means that there is the potential for substantial speed-up when doing many calls to the API, and using a persistent `Dataset` object instance.

It might be better to use this as an example, because a single call to `get` should be as fast via both APIs.

docs/basics/101-130-yodaproject.rst
```diff
@@ -0,0 +1,600 @@
...
+courses for statistical classification techniques in machine learning, and
+widely available, among many other sources as
+`GitHub gists <https://gist.github.com/netj/8836201>`_.
+
+Raw data as a modular, independent entity
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+To start your analysis project and comply to the YODA principles, you set up
+an independent data dataset with your project's raw data. For this, create a
+new dataset outside of ``DataLad-101``:
```
mih: I am still torn between the "create yourself a dataset" approach that you are taking here vs. us providing a dataset for this at https://github.com/datalad-handbook

Your approach is more self-contained, but cutting the prelude in favor of a plain `install` call removes some distractions and orthogonalizes the content with other chapters.

docs/basics/101-130-yodaproject.rst
```diff
@@ -0,0 +1,600 @@
...
+.. runrecord:: _examples/DL-101-130-102
+   :workdir: dl-101
+   :language: console
+
+   $ cd iris_data
+   $ datalad download-url https://gist.githubusercontent.com/netj/8836201/raw/6f9306ad21398ea43cba4f7d537619d0e07d5ae3/iris.csv
+
+.. todo::
+
+   or do we rather want to install/clone the dataset?
```
mih: Yeah... (see above)

docs/basics/101-130-yodaproject.rst
```diff
@@ -0,0 +1,600 @@
...
+sufficient provenance capture for my input dataset!" you think as you
+mentally tick of YODA principle number 2 for your input data.
+
+An analysis dataset
+^^^^^^^^^^^^^^^^^^^
+
+Now that you have this raw dataset, it is time to create an analysis for your
+midterm project.
+You start by creating an analysis dataset. Let's do it as a subdataset
+of ``DataLad-101``. Remember to specify the ``--dataset`` option within
```
mih: within -> of ?

docs/basics/101-130-yodaproject.rst
```diff
@@ -0,0 +1,600 @@
...
+(sepal and petal width and length) you will try to predict what type of
+flower (*Setosa*, *Versicolor*, or *Virginica*) a particular flower in the
+dataset is. You settle on two objectives for your analysis:
+
+#. Explore and plot the relationship between variables in the dataset and save
+   the resulting graphic as a first result.
+#. Perform a k-nearest neighbour classification on a subset of the dataset to
+   predict class membership (flower type) of samples in a left-out test set.
+   Your final result should be a statistical summary of this prediction.
+
```
mih: Maybe here, or right at the top of the chapter put a note that says something like "even if you don't know Python you can just c&p the code and follow along"

docs/basics/101-130-yodaproject.rst
```diff
@@ -0,0 +1,600 @@
...
+   :emphasize-lines: 8, 10, 13, 22, 41
+
+   $ cat << EOT > code/script.py
+
+   import pandas as pd
+   import seaborn as sns
+   from sklearn import model_selection
+   from sklearn.neighbors import KNeighborsClassifier
+   from sklearn.metrics import classification_report
+   from datalad.api import get
```
mih: I am leaning towards the approach proposed above and used right here for pandas as well: `import datalad.api as dl`....

docs/basics/101-130-yodaproject.rst
```diff
@@ -0,0 +1,600 @@
...
+
+   data = "input/iris.csv"
+
+   # make sure that the data is obtained (get will also install linked sub-ds!):
+   get(data)
```
mih: ... and `dl.get()` here. `get` is so generic, it is easy to miss the fact that this is a datalad command, especially for people not familiar with Python.

docs/basics/101-130-yodaproject.rst
```diff
@@ -0,0 +1,600 @@
...
+
+   The ``yoda`` procedure in ``midterm_project`` applied a different configuration
+   within ``.gitattributes`` than the ``text2git`` procedure did in ``DataLad-101``.
+   Within ``DataLad-101``, any text file is automatically stored in :term:`Git`.
+   This is not true in ``midterm_project``: Only the existing ``README.md`` files and
+   anything within ``code/`` are stored -- everything else will be annexed.
+   That means that if you create any other file, even text files, inside of
+   ``midterm_project`` (but not in ``code/``), it will be managed by :term:`Git-annex`
+   and content-locked after a :command:`datalad save` -- an inconvenience if it
+   would be a file that is small enough to be handled by Git.
```
mih: Maybe mention that such configuration scenarios can be combined: `create -c text2git -c yoda`

docs/basics/101-130-yodaproject.rst
```diff
@@ -0,0 +1,600 @@
...
+   GitHub is a web based hosting service for Git repositories. Among many
+   different other useful perks it adds features that allow collaboration on
+   Git repositories. `GitLab <https://about.gitlab.com/>`_ is a similar
+   service with highly similar features, but its source code is free and open,
+   whereas GitHub is a subsidiary of Microsoft.
+
+   Web-hosting services like GitHub and :term:`GitLab` integrate wonderfully with
+   DataLad. They are especially useful for making your dataset publicly available,
+   if you have figured out storage for your large files otherwise (as large content
+   can not be hosted by GitHub). You can make DataLad publish large file content to one location
```
mih: cannot be hosted **for free** -- they do offer hosting

docs/basics/101-130-yodaproject.rst
```diff
@@ -0,0 +1,600 @@
...
+     publish (ok: 2)
+
+.. gitusernote::
+
+   The :command:`datalad publish` uses ``git push``, and ``git annex copy`` under
+   the hood. Publication targets need to either be configured remote Git repositories,
+   or Git-annex special remotes (if they support data upload).
+
+Here is one important detail, though: By default, your tags will not be published.
+This requires an additional :command:`git push` with the ``--tags`` option to the
```
mih: There is a reason we that is the case. Tags are viral. Once pushed the propagate into any clone. However, especially the way a tag is used here, one may not actually want that. Let's say I remove the tag locally. Later on I add a clone of a friend as a remote in order to fetch a contribution. They still have the old tag, and BANG I remove my removed tag back where I no longer wanted it.

Maybe just skip the tag-push part here.

docs/basics/101-130-yodaproject.rst
```diff
@@ -0,0 +1,600 @@
...
+
+   Replace the ``url`` in the :command:`install` command below with the path
+   to your own ``midtermproject`` GitHub repository:
+
+   .. runrecord:: _examples/DL-101-130-119
+      :language: console
+      :workdir: dl-101/DataLad-101/midterm_project
+
+      $ cd ../../
+      $ datalad install -r --source "https://github.com/adswa/midtermproject.git"
```
mih: The recursive part is not needed. DataLad does that automatically on `rerun` based on the `input` specification we gave above. I know that this would make it impossible to keep the immediately following paragraphs as they are, but not having to know anything about the specific nature of the dataset in order to rerun IS a killer feature -- that is a bit shadowed by the present approach.

docs/basics/101-130-yodaproject.rst
```diff
@@ -0,0 +1,600 @@
...
+   on their creation was captured, we can simply recompute them with the
+   :command:`datalad rerun` command. Since we tagged the dataset once it was
+   ready for analysis, we can simply rerun any :command:`datalad run` command
+   since this tag:
+
+   .. runrecord:: _examples/DL-101-130-122
+      :language: console
+      :workdir: dl-101/midtermproject
+
+      $ datalad rerun --since ready4analysis
```
mih: If tags are not pushed above, this would need to be something else. Ultimately waiting for https://github.com/datalad/datalad/issues/3866

adswa:
Thanks much for the comments! They are really helpful!

docs/basics/101-130-yodaproject.rst
adswa: Good idea, thx

docs/basics/101-130-yodaproject.rst
adswa: I didn't even know this, thx

docs/basics/101-129-python.rst
```diff
@@ -0,0 +1,79 @@
...
+
+Todays lecture will give you an overview on DataLad's Python API and explore
+how to make use of it in an analysis project. Together with the previous
+section on the YODA principles, this section will be the basis for a data
+analysis midterm project.
+
+All of DataLad's user-oriented commands are exposed via ``datalad.api``.
+Thus, any command can be imported as a stand-alone command like this::
+
+   >>> from datalad.api import <COMMAND>
```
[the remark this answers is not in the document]
adswa: will do!

docs/basics/101-129-python.rst
adswa: Thanks for that! I'll change this. Thinking about this further, I think I will try to "downgrade" this section on Python to a subsection or hidden section. This way, the whole chapter will have a major focus on YODA (both in theory and practice), with some added benefit on learning about the Python API. This may also resolve the chaotic summary-situation (multiple summaries in one chapter) here, might be more intuitively structured, and shifts the focus from the Python API. No guarantee,  I'll see whether I can make this work.

docs/basics/101-130-yodaproject.rst
adswa: I'm indifferent - I'll push a dataset to https://github.com/datalad-handbook that can be installed as input and maybe do a hidden section about "if I have collected my own data..."

docs/basics/101-130-yodaproject.rst
adswa: Actually, I will add this to the chapter on configurations.

docs/basics/101-130-yodaproject.rst
adswa: For now I am back to rev-parsing the git log...

adswa:
This is restructured now, and I think this section works nicely as a practical session for YODA after the theoretical section. Few bits and pieces are yet unresolved, but I think its an improvement from the previous structure.

docs/basics/101-130-yodaproject.rst
```diff
@@ -0,0 +1,763 @@
...
+       :workdir: dl-101/DataLad-101/midterm_project
+       :realcommand: datalad --log-level critical siblings add -d . --name github --url https://github.com/adswa/midtermproject.git
+
+.. code-block:: bash
+
+   $ datalad create-sibling-github -d . midtermproject
+   .: github(-) [https://github.com/adswa/midtermproject.git (git)]
+   .: github(-) [https://github.com/adswa/midtermproject-input.git (git)]
+   'https://github.com/adswa/midtermproject.git' configured as sibling 'github' for <Dataset path=/home/me/dl-101/DataLad-101/midterm_project>
+   'https://github.com/adswa/midtermproject-input.git' configured as sibling 'github' for <Dataset path=/home/me/dl-101/DataLad-101/midterm_project/input>
```
mih: This is still the old output from recursive mode used before. Should be sufficient to kill the '-input' lines.

docs/basics/101-130-yodaproject.rst
```diff
@@ -0,0 +1,763 @@
...
+
+On GitHub, you will see a new, empty repository with the name
+``midtermproject``. However, the repository does not yet contain
+any of your dataset's history or files. This requires *publishing* the current
+state of the dataset to this sibling with the :command:`datalad publish`
+(:manpage:`datalad-publish` manual) command.
+
+.. code-block:: bash
+
+   $ datalad publish -r --to github
```
mih: No `-r` necessary anymore.

docs/basics/101-130-yodaproject.rst
adswa: thx for catching that!

adswa:
Tests passed, merging.

PR #231, opened 2019-10-30
add an example code list to Appendix

adswa:
This is how a "code list" looks ATM with the changes in autorunrecord proposed [here](https://github.com/mih/autorunrecord/pull/2).


mih:
Just a thought: Should we use this feature to get a more compact representation of the same information that is potentially more accessible in other contexts: http://docutils.sourceforge.net/docs/ref/rst/directives.html#list-table

There is also: http://docutils.sourceforge.net/docs/ref/rst/directives.html#csv-table (if this would be a better fit)

What do you think?

adswa:
IMO the table directive isn't exactly visually pleasing, and I suspect that with even smaller space for code (cells will be smaller than the full width of the page), the code snippets may actually become less readable. As for the .csv table, I'm not sure whether it adds much, or how we could have the code rendered nicely without having a weird looking (because .rst format containing) .csv file... that being said, it would be nice to have a more "portable" format for these collections of "follow-along code". I will continue thinking about this.

mih:
Look and feel could be tweaked with CSS, but I havent actually tried it...

adswa:
This is how it looks like (I have one in a WIP section on DataLad extensions):

image

mih:
OK, either way is fine with me.

adswa:
I'd suggest we'll checkout whether people are able to work with the snippets in the way they are displayed. Next weeks datamanagement course will get them for the first time (I yet have to create/add the snippets for the chapter on datalad run, though -- will put them into this PR).

adswa:
I have added the code snippets of session 2 in this PR as well.

adswa:
I'll merge this now to have it ready for tomorrows course -- we can see how well these work.

PR #235, opened 2019-11-04
remote analysis

### LINKED ISSUE #233: Use case: external scientist without data access needs results
This is a common problem for any data analysis involving personal information. Approach:

- Build dataset that implements the same structure (organization, and filenames if possible), but does not contain the actual problematic data (maybe tracked, but not available through annex, but maybe even without any relationship to the actual data, i.e. mock-data,or simulated data)
- Provide dataset publicly to aid development of analysis implementations
- Clearly describe how this mock differs from the inaccessible other dataset
- External users are instructed to create a new dataset (to hold their code) that has the mock dataset as a subdataset
- External users submit their dataset, the subdataset is replaced with the real dataset (actual version is tracked), code is executed (after having been reviewed), results are captured in the submitted dataset.
- Results are pushed back to the external users (or deposited in an accessible place for them to pull) -- the local data remains local and unavailable

adswa:
This wants to fix #233 when it's grown-up.
Currently, this only sketches the problem-space of the use case proposed in #233. A future step-by-step needs to detail the actual implementation.


docs/usecases/external_collaboration.rst
```diff
@@ -0,0 +1,364 @@
...
+The ABC-Lab collected data from patients for one of their studies.
+After publication, a Post-doc from the XYZ-Lab reaches out and proposes an interesting analysis of the same data on a new research question.
+Everyone involved is excited about this collaboration:
+The XYZ-Lab is happy that such a valuable dataset exists and is eager to give credit and co-authorship to collaborators in the ABC-Lab.
+The ABC-Lab is happy that someone else is interested in using their data and appreciates the possibility of another publication with lab-members as co-authors.
+
+However, the legal department in ABC's institute steps in: The data can not be shared with anyone outside as this would be a privacy violation.
+The data contains potentially identifying or personal information of the patients it was collected from.
+
+With such legal problems, the promising collaboration ceases before it even started, the data remains disclosed but also unused, and an interesting research question remains unanswered.
```
mih, line 27: I stumbled upon the phrase "disclosed but also unused". I am not sure what "disclosed" wants to communicate here? Maybe just

```suggestion
With such legal problems, the promising collaboration ceases before it even started, the data remains unused, and an interesting research question unanswered.
```

docs/usecases/external_collaboration.rst
```diff
@@ -0,0 +1,364 @@
...
+
+The DataLad Approach
+^^^^^^^^^^^^^^^^^^^^
+
+Scientists take their participants' privacy very seriously, and everyone is well aware that certain data can not leave the institute's computational premises.
+Nevertheless, valuable data should not remain accessible to only selected few.
```
mih, line 33:
```suggestion
Nevertheless, valuable data can be made accessible without providing direct access.
```

Cut moral imperative and focus on technical aspect.

docs/usecases/external_collaboration.rst
```diff
@@ -0,0 +1,364 @@
...
+To enable data analysis by others without disclosing private information researchers can make use of DataLad.
+
+In order to *develop* analysis scripts that can analyze the data, the actual sensitive data is not needed.
+A mock dataset that has the same structural layout and the same file names as the original dataset can already be sufficient information to create a working script.
```
mih, line 37: This is one way to do this, but there is also another:

Instead of a completely separate mock dataset, it could be the actual dataset with some, clearly identified, mock data points. This has the advantage that an external consumer and precisely indicate which version of a dataset they want to ingest.

This works less well for something like tabular data, where all information is in a single file. But it is pretty simple in something like imaging dataset, where on an image-by-image, file-by-file basis the data availability can be decided on the basis of annex key accessibility.

Not sure whether it is worth putting this complexity here. But it should be possible and meaningful to point out this flexibility.

docs/usecases/external_collaboration.rst
```diff
@@ -0,0 +1,364 @@
...
+This ensures that the script is adjusted to the actual file names and works "out-of-the-box".
+
+Afterwards, this dataset is given back to the ABC-lab.
+They can review the provided code and exchange the mock dataset with the original, sensitive data.
+With this, they can run the code provided by the collaborator such that sensitive data stays within their institute.
+All results computed from this are aggregated into their collaborator's dataset.
+Upon completion, only the results, not the sensitive input data, are pushed back.
+
+
+.. find-out-more:: Advice for the holder of sensitive data
```
mih, line 52:
```suggestion
.. find-out-more:: Advice for the controller of sensitive data
```

I am proposing to adopt GDPR-style terminology. I am under the impression that a "holder" here is used in the same fashion as a "data controller" in GDPR. Using the more standard term should improve understanding.

docs/usecases/external_collaboration.rst
mih: I now see that this is pretty much already covered further down. So please take my comment as a "possibly label the two alternatives more clearly already here".

docs/usecases/external_collaboration.rst
```diff
@@ -0,0 +1,364 @@
...
+.. runrecord:: _examples/remote-analysis-107
+   :language: console
+   :workdir: usecases/remote-analysis/penguin-jelly
+
+   $ git tag runme
+
+The question is: How does this look like on the real data?
+A data provider would clone your dataset and replace the subdataset with the actual sensitive dataset.
+
+We can pretend to do this:
```
mih, line 291: I understand why the clone of the external collaborator is reused here. But I think this should not be done, because it makes it hard to see where exactly the line between provider and consumer is.

I think the `drop` should go and this part should start with a section heading (Provider perspective) and a `clone` from the source where the prepared YODA dataset can be found. This could either be the one prepared in the code above, or a dedicated "edu" resource. Given that we already have to have `empty` and `mock`, we can also have `jelly` hosted there.

This would make the message "put it somewhere accessible for the provider".

docs/usecases/external_collaboration.rst
mih: I now see that "provider" is used frequently further down. This label is also clearer than "holder" and could be used here too.

docs/usecases/external_collaboration.rst
mih: OK, I now see that this section heading comes next already.

I find this confusing. The rerun demo here appears to communicate the provider perspective to the consumer. But the next section communicates the provider perspective to any reader. Hence this feels duplicate and adds unnecessary complexity.

docs/usecases/external_collaboration.rst
```diff
@@ -0,0 +1,364 @@
...
+Then, we rerun the analysis:
+
+.. runrecord:: _examples/remote-analysis-109
+   :language: console
+   :workdir: usecases/remote-analysis/penguin-jelly
+
+   $ datalad rerun runme
+
+The data providers perspective
+""""""""""""""""""""""""""""""
```
mih, line 310:
```suggestion
The data provider perspective
"""""""""""""""""""""""""""""
```

docs/usecases/external_collaboration.rst
```diff
@@ -0,0 +1,364 @@
...
+
+How does it look like from the data provider perspective?
+A possible first step could be the initial creation of a DataLad dataset.
+To be able to keep all or selected file contents private, data providers should make sure that
```
mih, line 314: OK, this is about preparing the dataset that has already been consumed in the section immediately above.

I find this confusing, because it jumps through time.

Maybe it would be better to split the "toy example" into three sections that are ordered chronologically:

1. data provider: prepare a dataset for remote analysis
2. data consumer: prepare a remote analysis
3. data provider: execute a remote analysis

docs/usecases/external_collaboration.rst
adswa: addressed by reordering as you proposed.

docs/usecases/external_collaboration.rst
```diff
@@ -0,0 +1,429 @@
...
+- or clone a dataset with simulated file contents: https://hub.datalad.org/edu/penguins-mock.git
+
+.. runrecord:: _examples/remote-analysis-103
+   :language: console
+   :workdir: usecases/remote-analysis/penguin-jelly
+
+   $ datalad clone -d . https://hub.datalad.org/edu/penguins-mock.git inputs
+
+Afterwards, it is time to "develop" your analysis.
+In this toy example, you can download it, but its also a good exercise to take a look at the available information in the shared dataset and write a script to compute something simple from it - like for example a mean.
```
jennaswa, line 228:
```suggestion
In this toy example, you can download a demo analysis, but its also a good exercise to take a look at the available information in the shared dataset and write a script to compute something simple from it - like for example a mean.
```

docs/usecases/external_collaboration.rst
```diff
@@ -0,0 +1,429 @@
...
+
+      # fit the model on data; predict Species of own data points
+      model.fit(data, target)
+      res = model.predict(local_samples)
+      local_samples['prediction'] = res
+      # save the prediction
+      local_samples.to_csv(args.output)
+
+
+To make things easier for the data provider, you can add a software container that includes all required software, using the ``datalad-containers`` extension.
```
jennaswa, line 358: This is another example where it might help the reader navigate and go back to this as a resource by adding additional heading levels to the different "Perspective" sections.

docs/usecases/external_collaboration.rst
```diff
@@ -0,0 +1,429 @@
...
+
+.. runrecord:: _examples/remote-analysis-105
+   :language: console
+   :workdir: usecases/remote-analysis/penguin-jelly
+
+   $ datalad containers-add software --url shub://adswa/resources:2
+
+Let's run the script on mock data in the software container.
+The results will not be correct because the data is only simulated, but it generates a run record that the data provider can very easily rerun - sparing you the need for detailed instructions in an Email or a README file.
```
jennaswa, line 367:
```suggestion
The results will not be correct because the data is only simulated, but it generates a run record that the data provider can very easily rerun - sparing you the need for detailed instructions in an email or a README file.
```

docs/usecases/external_collaboration.rst
```diff
@@ -0,0 +1,429 @@
...
+   :workdir: usecases/remote-analysis/penguin-jelly
+
+   $ git tag runme
+
+
+
+3. Perspective: Execute the remote analysis
+"""""""""""""""""""""""""""""""""""""""""""
+
+The question is: How does this look like on the real data?
```
jennaswa, line 390:
```suggestion
The question is: How does this look on the real data?
```

docs/usecases/external_collaboration.rst
```diff
@@ -0,0 +1,429 @@
...
+A data provider would clone your dataset and replace the subdataset with the actual sensitive dataset.
```
jennaswa, line 391:
```suggestion
A data provider would clone the external collaborator's dataset and replace the subdataset with the actual sensitive dataset.
```

docs/usecases/external_collaboration.rst
```diff
@@ -0,0 +1,429 @@
...
+
+.. runrecord:: _examples/remote-analysis-108
+   :language: console
+   :workdir: usecases/remote-analysis/penguin-jelly
+
+   $ datalad drop --what datasets inputs
+   $ datalad clone -d . https://hub.datalad.org/edu/penguins.git inputs
+
+
+Then, we rerun the analysis:
```
jennaswa, line 403:
```suggestion
Then, the data provider would rerun the analysis:
```

docs/usecases/external_collaboration.rst
```diff
@@ -0,0 +1,429 @@
...
+
+.. runrecord:: _examples/remote-analysis-109
+   :language: console
+   :workdir: usecases/remote-analysis/penguin-jelly
+
+   $ datalad rerun runme
+
+
```
jennaswa, line 411:
```suggestion
Finally, the data provider would share the resulting outputs back to the external collaborator.
```

jennaswa: Or something to conclude with what the sharing back process would look like.

docs/usecases/external_collaboration.rst
adswa: thanks for the suggestion. Just as before there is the issue that the alternative suggestion is at the moment not doable (at least by default). People would lack the software environment. I'm reluctant to add software installation instructions, as they would make this feel much more complex. The fix would be a script that has fewer software requirements, e.g., bash-only, but this would be a trade off against the "interestingness"/"life-likeness" of the example - I can only come up with very crude data science in bash

adswa:
Thanks all for the extensive review this has gotten. I've pushed a few final changes and will merge once CI has turned green.

PR #244, opened 2019-11-05
Remove contractions for improved accessibility

### LINKED ISSUE #243: Replace all contractions in the book
A while ago I stumbled across a tweet that mentioned how it is much easier to differentiate "do" and "do not" than "do" and "don't" for people with reading disabilities/blindness etc that rely on software reading books aloud for them.
Let's take care of replacing all contractions in the book with the full length expression.

adswa:
This fixes #243.

### DIFF
--- a/docs/404.rst
+++ b/docs/404.rst
@@ -4,7 +4,7 @@
 404 — Not Found
 ###############
 
-**Sorry, but we couldn't find the page you requested.**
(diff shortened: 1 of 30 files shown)


mih:
Looks good, but conflicts ATM.

adswa:
Ah, thx, didn't notice -- will fix. :)

PR #248, opened 2019-11-09
if you do know git ... and if you dont

jbpoline:
seems that the alternative was : if you dont know git, ... otherwise, if you dont know git ... :)

### DIFF
--- a/docs/intro/philosophy.rst
+++ b/docs/intro/philosophy.rst
@@ -114,7 +114,7 @@ that captures the spirit of what DataLad is, and here is a brief overview on it.
 #. **A dataset is a Git repository**. If you know the :term:`version control` system
    :term:`Git`, and what you can use it for,
    then take notice that everything you can use Git for also applies to everything managed
-   by DataLad -- plus much more. If you don't know Git
+   by DataLad -- plus much more. If you do know Git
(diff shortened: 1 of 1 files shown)


adswa:
Hi JB! :)

Maybe I'm reading it wrong, but the sentence is (roughly):
If you know Git --> cool, If you don't know Git --> don't worry

So the "don't" should indeed be correct there. However, I think this PR hints at that these sentences are a bit unclear in the way they are currently phrased...
Maybe we should get rid of the "do know - don't know" style and go with something simpler & restructured like this:

> A dataset is a Git repository. All features of the :term:`version control` system :term:`Git` also apply to everything managed by DataLad – plus many more.
If you don’t know or use Git yet, there is no need to panic – there is no necessity to immediately learn all of Git to follow along in learning and using DataLad. You will experience much of Git working its magic underneath the hood when you use DataLad, and will soon start to appreciate its features.


Do you think this makes it a bit clearer?

(Warm wishes and a hug from Germany!)

jbpoline:
sounds good ! see my update !
warm hello and hug as well !

adswa:
Cool, thanks!
Hey @all-contributors, please add @jbpoline for content!

PR #254, opened 2019-11-12
ecompute all examples, fixes #252

### LINKED ISSUE #252: Out-of-sync Commit hashes
In some examples, the displayed commit hashes and dataset id's come from different histories, see e.g. mismatch in log and commit that is rerun [here](http://handbook.datalad.org/en/latest/basics/101-109-rerun.html). We'll need to rebuild all of the examples once soon, I'd suggest.

adswa:
This fixes #252 -- at least temporarily ;-) I'd suggest that we should recompute all examples whenever a major addition happens, for example when section is added .

I realized when recomputing all examples that the sections on "back and forth in time" and "misc filesystem operations" where not yet correctly numbered, and fixed this as well.


mih:
Cool, thx!

PR #263, opened 2019-11-13
add paragraph on placeholders to datalad run

adswa:


adswa:
I've now also added a paragraph on escaping curly brace characters while I was at it.

PR #268, opened 2019-11-14
justify text alignment

adswa:
This will justify all text in the handbooks html version. Its a matter of personal taste, what do the others think about this?


adswa:
image

mih:
Big fan! I'd say, let's do this until we find solid evidence against it.

PR #274, opened 2019-11-26
some updating after linting for insensitive words

adswa:
(have only done the first half of the book, the other will come soon)

### DIFF
--- a/docs/basics/101-101-create.rst
+++ b/docs/basics/101-101-create.rst
@@ -122,5 +122,5 @@ Let us now put some content inside.
 
 .. rubric:: Footnotes
 
-.. [#f1] A nice and easy tool we can recommend as an alternative to :command:`git log` is :term:`tig`.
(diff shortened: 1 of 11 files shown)


adswa:
This fixes #275. According to the alex.js tool, the book is quite profane. I'm confident that words like "color", "execution", or "host" are not insensitive given their context in the book (and that there is no useful equivalent/synonym). I'm am unsure how my heavy use of "easy" or "simple" affects readers. I'm using them when something is (hopefully objectively) simple, or to "convince" readers that a seemingly difficult task is indeed easier than thought, or in direct speech of the book characters. I suspect that these words can be infuriating if one reads a section and finds the contents incredibly difficult. I hope this does not happen, but in that I somewhat rely on user-feedback. I've decided that until someone complains about the book being condescending I'll stick to my writing style, but I'm happy to reconsider my take on those words later.

mih:
I support your decision.

Issue #38, opened 2019-07-01
Adopt DataLad colors


mih:
ATM the theme sphinx uses for HTML and PDF output is using the default colors. However, even now there is a figure included that uses the color scheme we have adopted for DataLad elsewhere (website, posters https://f1000research.com/posters/7-1965).

I would make sense to me to adopt the same colors also for the book (notes, todos, highlights, etc.). Let me know, and I'll give it a go.

adswa:
I think that is a very good idea. tbh I've been interestly eying other sphinx themes just because the have a more datalad-like color scheme -- but if we can get the DataLad color into this theme, even better!

mih:
The HTML version already has some of this now. The PDF needs more attention.

Latex color setup is detailed here: http://www.sphinx-doc.org/en/master/latex.html#sphinxsetup-key

adswa:
I'm adding a few more issues that the PDF has here. I don't think that they are urgent, but I believe they must be annoying or confusing if one reads only the PDF:

- [ ] ``findoutmore`` directories are expanded, but not distinguishable in any way from the rest of the content. It would be cool if we could at least give them another background color. As far as I can see, this doesn't work with any of the [keys](https://www.sphinx-doc.org/en/master/latex.html) that Sphinx provides for the Latex customization
- [ ] The same is true for ``gitusernotes``
- [ ] The PDF labels book parts differently from how we refer to them in the text. Every ``.rst`` file is a "chapter" (we call them sections), hence we currently have 58 chapters. The next-level grouping into what we call "chapters" (e.g. "DataLad Datasets" with sections 101-101-create until 101-107-summary) are called "part" in the PDF, hence we currently have 15 parts. One of those parts is the Introduction, 1 part is the usecases, and 9 parts are currently the Basics. I don't see how we could configure the PDF to be in line with how we call the different parts. One solution could be to add a "If you are reading the PDF version, be mindful..." paragraph to the "How to read the handbook" section. This could be done with ``.. only:: latexpdf`` to have it not rendered in the html version even. Alternatively, we may need to reconsider our wording and adjust it towards the Latex naming.
- [ ] the cheatsheet is too large for a single PDF page. Instead of previewing the cheatsheet in full, we could include one figure per cheat sheet page, so that the cheat sheet are several successive pages in the PDF
- [ ] sections that are "hidden" via 
```
.. toctree::
   :hidden:
```
  actually do show up in the book. Its not the case for ``.. only:: adminmode`` parts
[...]

PR #292, opened 2019-12-03
DataLad Colors

adswa:
This inches towards #38. A main deficiency of the PDF is that the colored admonition boxes are not preserved. Instead, notes or other directives that are not codeblocks are simply distinguished from normal text by a bit of spacing and table lines(?). This adds almost-DataLad-colored boxes around Note and Important directives (haven't figured out a general Admonition role yet...)

**Before**:

image

**After**:

image

This needs further work (e.g., also TODOs are now brightly orange), but its a start

### DIFF
--- a/docs/_static/custom.css
+++ b/docs/_static/custom.css
@@ -39,8 +39,9 @@ div.admonition-todo {
  * idea is to clearly set them apart from the normal
  * content, because they should be completely optional */
 div.gitusernote {
-    background-color: #c3ebffff;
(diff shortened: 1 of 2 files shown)


adswa:
More playing around with colored boxes (in html): trying to adjust findoutmore to DataLad colors and Gitusernotes to Git colors...

**before**:

image


**after**:

image

adswa:
With orange progress bars we're getting closer to a standardized color scheme

image

PR #296, opened 2019-12-07
Minor sentence and typo fixing

nhjjreuter:
The section ends with the paragraph "_But enough of the abstract talking. You came here to learn, and the handbook will not waste your time further by requiring you to only read – let’s start to use DataLad._"

However, the next section is "How to use the handbook". I suggest removing this pargraph, or moving it to the "How to use the handbook" section.

### DIFF
--- a/docs/intro/philosophy.rst
+++ b/docs/intro/philosophy.rst
@@ -151,7 +151,7 @@ that captures the spirit of what DataLad is, and here is a brief overview on it.
 
 #. Simultaneously, though, DataLad aims to
    **maximize the (re-)use of existing 3rd-party data resources and infrastructure**.
-   Users *can* use an existing central infrastructure should they want to.
+   Users *can* use existing central infrastructures should they want to.
    DataLad works with any infrastructure from :term:`GitHub` to
    `Dropbox <https://www.dropbox.com>`_, `Figshare <https://figshare.com/>`_
    or institutional repositories,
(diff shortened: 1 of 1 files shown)


docs/intro/philosophy.rst
```diff
@@ -136,22 +136,22 @@ that captures the spirit of what DataLad is, and here is a brief overview on it.
...
-   institutions webserver, or any other common computational infrastructure you
+   way, no central infrastructure needs to be maintained (or paid for).
+   Your own laptop is the perfect place for your DataLad project to live, as is your
+   institution's webserver, or any other common computational infrastructure you
    might be using.
 
 #. Simultaneously, though, DataLad aims to
    **maximize the (re-)use of existing 3rd-party data resources and infrastructure**.
-   Users *can* use existing central infrastructure should they want to.
+   Users *can* use an existing central infrastructure should they want to.
```
adswa: Sorry for being nit-picky, but I don't want to create the impression that *one* infrastructure needs to be chosen (any number would be fine)

```suggestion
   Users *can* use existing central infrastructure should they want to.
```

nhjjreuter: In that case I'd suggest changing it to "Users *can* use existing central infrastructures should they want to." Would that fit better?

adswa: sounds good!

Issue #124, opened 2019-08-25
Add acknowledgements section


mih:
People, institutions, funders, grant numbers, logos, etc.

mih:
The 'people' aspect will be addressed by #140

adswa:
Some of the funding was mentioned with #308, but there is more funding to be mentioned, right?

adswa:
Will have an up-to-date overview in the last minor tune ups for the 0.13 release. closing for now.

PR #308, opened 2019-12-11
Acknowledgements

adswa:
I'm addressing some aspects of #124 by copying the acknowledgements from the datalad website.


mih:
I'll merge this as-is, but we should update it with the latest info soonish.

PR #317, opened 2019-12-12
Update .zenodo.json

marisaheckner:


adswa:
Thx! Do you have an ORCID ID you want to add as well?

marisaheckner:
Not yet, does it make sense to register?
Btw nicely written abstract. I'm struggling with mine atm...

adswa:
Yes, its very useful to have an ORCID! I'd recommend it (and the library here requires it actually). Its fast, checkout [orcid.org](https://orcid.org/). And its good to have your ORCID here, because once well release the books next version (soon), that will be directly credited to you.

marisaheckner:
Oh I apparently signed up before ;-) I added it, thanks!

adswa:
awesome, thx!

PR #319, opened 2019-12-12
ENH: added zenodo entry for myself, unified some affiliations, added access_right field

yarikoptic:
Note that adding access_right open also makes "license" field required
https://developers.zenodo.org/#representation


adswa:
Cool, thx! As "license" field is already in, I don't think I need to add anything more, right?

PR #335, opened 2020-01-09
always append mih as last author to list of authors

adswa:
The change in this PR ensures that @mih will always stay the last author of the book in the PDF version.


mih:
I hope I'll soon be on page 5 of the authors list ;-)

PR #337, opened 2020-01-09
remove rc specification in installation

adswa:
This is to be merged once DataLad [``0.12.0``](https://github.com/datalad/datalad/pull/3993) is released.


mih:
It might actually be helpful to keep the version specs in the `pip install` commands. It would benefit users that have datalad 0.11 installed, because without them an upgrade may not necessarily happen.

adswa:
True, thx! Done.

PR #342, opened 2020-01-10
Add a changelog

adswa:
I don't think its paticularly useful or necessary to have a change log for a book, but I wanted one anyway for good practice. :)


mih:
I think it is useful. Especially because your writing pace is so high. People who have read some version before will benefit from a high-level idea of what is worth looking at (again).

PR #353, opened 2020-01-14
Further Latex improvements

adswa:
- ~remove URLs in Footnotes. We can rely on the href'ed text~
- Colored subheadings in more DataLad-dy color scheme, code boxes. TODO directives are not brightly orange anymore, but framed between slim orange bars
- Enforce figure positions in the text. This places the figures where they need to be, and prevents empty pages surrounding a single figure on an additional otherwise empty page.
- remove blank pages after toctrees and after headings (-> 50 pages fewer).

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -278,7 +278,7 @@ latex_show_urls = 'footnote'
 latex_elements = {
     'papersize': 'a4',
     'pointsize': '11pt',
-    'figure_align': 'tbp',
+    'figure_align': 'H',
     'fncychap' : r'\usepackage[Bjarne]{fncychap}',
     'passoptionstopackages': r'\PassOptionsToPackage{svgnames}{xcolor}',
     'preamble': r"""
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -274,7 +274,7 @@ latex_logo = None
 
 latex_toplevel_sectioning = 'part'
 latex_show_pagerefs = True
-#latex_show_urls = 'footnote'
+latex_show_urls = 'footnote'
 latex_elements = {
     'papersize': 'a4',
     'pointsize': '11pt',
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -279,6 +279,7 @@ latex_elements = {
     'papersize': 'a4',
     'pointsize': '11pt',
     'figure_align': 'H',
+    'extraclassoptions': 'openany,oneside',
     'fncychap' : r'\usepackage[Bjarne]{fncychap}',
     'passoptionstopackages': r'\PassOptionsToPackage{svgnames}{xcolor}',
     'preamble': r"""
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -279,24 +279,14 @@ latex_elements = {
     'papersize': 'a4',
     'pointsize': '11pt',
     'figure_align': 'tbp',
+    'fncychap' : r'\usepackage[Bjarne]{fncychap}',
+    'passoptionstopackages': r'\PassOptionsToPackage{svgnames}{xcolor}',
     'preamble': r"""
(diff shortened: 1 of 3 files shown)


mih:
How would URLs be accessible in a printed PDF? The printed PDF is the primary, if not only purpose of the PDF from my POV.

adswa:
Oh true :(

docs/conf.py
```diff
@@ -278,25 +278,16 @@
 latex_elements = {
     'papersize': 'a4',
     'pointsize': '11pt',
-    'figure_align': 'tbp',
+    'figure_align': 'H',
+    'extraclassoptions': 'openany,oneside',
```
mih: In a real book, it is rather common to have chapters only start on the right. But I am fine with this change.

docs/conf.py
```diff
@@ -278,25 +278,16 @@
 latex_elements = {
     'papersize': 'a4',
     'pointsize': '11pt',
-    'figure_align': 'tbp',
+    'figure_align': 'H',
```
mih: My body stopped twitching now....

In my experience, the desire to force the placement of figures in a PDF is almost always in indication of suboptimal figure captions. A document often suffers substantially in readability when figures are not placed into the normal locations: at the top, the bottom, or covering a full page.

adswa: I made the change after trying unsuccessfully to make it work in another way. This for example is what the previous configuration did to the "what you really need to know" part (left: old, right: with the change in this PR) - it just puts 4 of the images right after another, out of the context of the small text box it applies to:

image

Here is another one (again, left = old, right = with change in this PR): there are two screenshots, on the right you can see that the text starting with "The above screenshot" should be below the screenshot, and is helpful to visually distinguish the two screenshots (on the right), but it currently is above the screenshots.

image

I could indeed make use of subtitles for the figures to fix this, but I don't think this would guarantee correct figure placement in all cases.

docs/conf.py
adswa: I know its useful for printing, but IMHO I'm always confused when I scroll through the PDF on my computer and have up to two consecutive blank pages. I'm not sure whether the majority of PDF consumers print this -- my (completely based on thin air) assumption is that they consume it as a digital document rather than printing. But if you think otherwise, lets keep it friendlier for printing.
On an even more subjective level, I find 346 pages (without the blank ones) less intimidating than 390 pages (with the blank ones).

docs/conf.py
adswa: I actually scrolled through the whole PDF after your comment, and only then I really noticed how many figures are funnily messed up with the current configuration. In any suboptimal case, the "H" configuration improves the visual experience and the comprehensiveness of the contents, and there is no figure where it makes things worse.
Obviously, there are still plenty of problems with the figures. Some figures resolution is really low, and the sizing of many figures seems to be off. 

If you have ideas on how to improve on the current state of figure placement I'm happy to hear them, but I think this small change really goes a long way for very little effort.

docs/conf.py
mih: We can (later) address that be rendering multiple versions of PDF -- for printing and electronic consumption. I am perfectly happy with this change.

docs/conf.py
mih: OK :raised_hands: ;-)

Let's assume I never said anything ;-)

(we will never be perfect with an autogenerated PDF, and I will now go and sacrifice three oxen to Donald Knuth ... that should get me out of the guilt mode)

adswa: I'm so sorry for the pain this inflicts, and my blasphemy ;-) you can always blame me for this ;-)

mih: ;-) I will survive. The PDF is objectively better than before with this change applied.

PR #357, opened 2020-01-15
restructure and fix zenodo

adswa:
- There were two commas that invalidated the json file
- Removed creators versus contributors distinction to put mih at the bottom to be last author

[jsonlint](https://jsonlint.com/) says its valid json, and authorship is fixed to Wagner, A. S. ... & Hanke, M.


adswa:
I'm merging, and will then create a signed tag on master and archive to zenodo afterwards

PR #361, opened 2020-01-16
Cheatsheetfix

### LINKED ISSUE #360: cheatsheet mismatch between website and github
The cheatsheet .svg on the [website](http://handbook.datalad.org/en/latest/basics/101-136-cheatsheet.html) has misaligned titles and blank sections (top-right), but the version on GitHub .pdf doesn't. It looks like the src on the website just needs to be updated to the GitHub version.

adswa:
This fixes #360. 
I have pushed updates to the submodules Makefile that render a separate, page-wise plain .svg version with all text being converted to paths (to prevent browser-dependent appearance of the fonts used in the .svg). With this PR, the cheatsheet preview links to this new file.


adswa:
I think this is uncontroversial from the side of the handbook - will merge to have a nice preview.

PR #366, opened 2020-01-20
Restructure usecases

adswa:
Whether or not we will go forward with https://github.com/datalad/datalad/pull/4046#issuecomment-575920252, this restructures the toctree:

- Use cases are numbered (I went for roman numbers, but that's an arbitrary choice made for slightly better distinction to the Basics numbering) and get short, general headings
- Each use case gets a short description in the toctree
- Small figures separate major parts of the book (Intro-Basics, Basics-Usecases)

I think this makes the use cases more visible. 

TODO:

- [x] fixup PDF version.

### DIFF
--- a/docs/contents.rst.inc
+++ b/docs/contents.rst.inc
@@ -12,18 +12,28 @@
    intro/howto
    intro/executive_summary
 
-.. figure:: artwork/src/basics.svg
-   :width: 60%
-   :target: http://handbook.datalad.org/en/latest/basics/intro.html
+.. only:: latex
+
+   ##########
+   **Basics**
+   ##########
+
+.. only:: html
+
+   .. figure:: artwork/src/basics.svg
+      :width: 60%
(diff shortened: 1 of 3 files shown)


adswa:
This PR simultaneously also improves the PDF version of the book by removing toctree overview of use cases, and introducing a PDF only "Use cases" heading that makes navigating to the use cases in the PDF easier.

mih:
My 2ct: I envision the number of use cases to grow quite high in the future. Going for roman numbers will make us face MCDXXVIII at some point ;-)

adswa:
arggg... this PR is introducing some ugly conflicts in #367... I should have marked this one to be not merged...

mih:
Do you want to revert the merge. I am ok with that, and/or can do it

adswa:
I tried, but unsuccessful, and then resolved conflicts in #367 in the end

PR #371, opened 2020-01-27
reset sibling configuration properly

adswa:
previously, the git reset was attempted in a file in .git. I just noticed it while playing around with `update` in the created dataset.


adswa:
I'm merging this now.

PR #373, opened 2020-01-28
The dataset argument in python functions

adswa:
After exploring the difference in path interpretations across Python functions I thought it would be useful  briefly mention how information about the dataset a command should operate on can be passed to Python functions.

I'm also adding a note that using dataset methods or specifying the ``dataset=`` argument with a Dataset instance (as opposed to a string) yields more consistent behavior.

### DIFF
--- a/docs/basics/101-130-yodaproject.rst
+++ b/docs/basics/101-130-yodaproject.rst
@@ -82,7 +82,7 @@ section below:
     As shown above, the only required parameter for a Dataset is the ``path`` to
     its location, and this location may or may not exist yet.
 
-    Stand-alone functions have a ``dataset=`` argument, similar to the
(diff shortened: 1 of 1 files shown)


docs/basics/101-130-yodaproject.rst
```diff
@@ -82,6 +82,31 @@ section below:
     As shown above, the only required parameter for a Dataset is the ``path`` to
     its location, and this location may or may not exist yet.
 
+    Stand-alone functions have a ``dataset=`` argument, similar to the
```
mih: s/similar/corresponding/ to ...

adswa: thx!

adswa:
I assume this is good to go ;-) Will merge

PR #377, opened 2020-02-05
MNT: add more index entries throughout the book

adswa:


adswa:
Updated this, and added some more intersphinx mapping references for the datalad docs (https://github.com/datalad/datalad/pull/4046)

PR #380, opened 2020-02-07
Pin Sphinx version for readthedocs to prevent #379

### LINKED ISSUE #379: Broken meta data in HTML after #367?
It appears as if #367 (it introduced section numbering) also introduced a meta property HTML tag that seems to be broken and results in this:

image

adswa:
ATM we don't have a readthedocs.yml file, and Readthedocs builds the handbook with the most recent Sphinx release (2.3.1). I believe that this version leads to a tiny bug, described in #379. By pinning Sphinx to an earlier version in ``requirements.txt`` and providing a ``readthedocs.yml`` file, I hope to get rid of the broken meta property tag.

Hopefully fixes #379.

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -1,0 +1,23 @@
+# .readthedocs.yml
+# Read the Docs configuration file
+# See https://docs.readthedocs.io/en/stable/config-file/v2.html for details
+
+# Required
+version: 2
+
+# Build documentation in the docs/ directory with Sphinx
+sphinx:
+  configuration: docs/conf.py
+
+# Build documentation with MkDocs
+#mkdocs:
+#  configuration: mkdocs.yml
+
+# Optionally build your docs in additional formats such as PDF and ePub
+formats: all
+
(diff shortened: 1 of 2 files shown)


adswa:
An alternative "fix" would be to remove the line in the layout template that suffers from the change in the most recent Sphinx version: https://github.com/datalad-handbook/book/blob/6be555db8061a8e2e3abd7f08db7c1cf8aaa28f2/docs/_templates/layout.html#L31

(the problem is the "title" template -- it includes a <span class="section-number">1. </span> around the section number now, which is what breaks things)

adswa:
I'll see whether this yaml file works. Its recommended by rdt to have a yml file anyway, and I want the broken tag to not be displayed on every html page anymore.

adswa:
Wonderful, this worked.

adswa:
Ha, I take this back. The builds currently fail because requirements.txt does not specify `dataladhandbook_support` (which is to be installed with `pip install -e .` in the repo). I need to find out how to specify this in requirements or revert this and get rid of the meta tag in the template...

PR #383, opened 2020-02-12
UX: copybutton for easier copying of code snippets

adswa:
The copybutton Sphinx extension introduced a feature to specify text (such as leading `$` in our code snippets) to be explicitly not copied. With this, 90% of code snippets are easy to copy. It adds a small copy symbol to the snippet and copies the code (no command output) to the clipboard.

It works for multiline-commands (e.g., ``run``), but unfortunately not for any of the here documents. Especially for the codelists, though, I envision this feature to be helpful.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -79,8 +79,10 @@ extensions = [
     'sphinxcontrib.rsvgconverter',
     'sphinxcontrib.plantuml',
     'dataladhandbook_support',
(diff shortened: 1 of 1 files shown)


adswa:
Ah sorry, nevermind. The released version of copybutton does not do this - I didn't realize I had a custom installation with own code changes. In that case, I'll try to improve upon mine and get rid of the last inconveniences.

PR #396, opened 2020-02-20
ENH: extensions - add crawler and mention that they might be available "natively"

yarikoptic:
- crawler (the cradle of DataLad AFAIK) was left behind :-(  It could be an ugly old duckling ATM, but IMHO still worth mentioning.
- `pip` is one way to install.  
Installation chapter already introduces to `conda` etc.  So I think it is worthwhile at least mentioning that extensions might also be available from originally used distribution.
I also expect https://github.com/datalad/datalad-extensions (or later some gh-pages view of it) will be improved to provide an overview of their status etc.

### DIFF
--- a/docs/basics/101-144-intro_extensions.rst
+++ b/docs/basics/101-144-intro_extensions.rst
@@ -70,4 +70,11 @@ such as in
 
    $ pip install datalad-container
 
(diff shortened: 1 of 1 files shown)


adswa:
Thx @yarikoptic! I've pushed a link to the glossary term of datasets.datalad.org on top of this.

PR #399, opened 2020-02-22
BF: attempt to specify handbook support in requirements.txt

adswa:
This tries to fix #398 (not sure whether this works)

PR #401, opened 2020-02-22
bf/tmp: explicitly include submodules.

adswa:
I suspect a previous failure in #399 was due to this


adswa:
cool, this seems to fix it.

PR #403, opened 2020-03-01
Towards user type specific guides to the book

adswa:
This is an idea from the Turing way: Identify with a "persona" (one of several stereotypical users) and get pointers to specific parts of the book. I'm not sure whether this idea will be feasible in the end, but here is a start to explore how it may look.

The document is hidden in the toctree, and only appears in the introduction on the index page.


docs/intro/user_types.rst
```diff
@@ -0,0 +1,123 @@
...
+:ref:`symlink` for some Basics on :term:`git-annex`.
+
+If you're seeking help with regard to large datasets, checkout section
+
+.. todo::
+
+   write section on scaling up
+
+If none of this helps, don't hesitate to file an issue or post your question
+to `Neurostars.org <https://neurostars.org>`_. We are happy to help!
```
yarikoptic: I think might be worth to explicitly point to "Search" on the handbook page.  Also, since "ran into a problem", it is not unlikely that user hit a bug.  Search through bug reports on github would be useful but "help" section refers only to neurostars (it seems nowhere in the book there is a reference to https://github.com/datalad/datalad/issues on purpose or not). Situation is a bit tricky since issues could be due to extensions, but at least searching through datalad organization could lead to an answer. 

Also, related, hopefully eventually https://github.com/datalad/datalad/issues/3942 would be pushed forward, so users running into the bugs would get (open or closed) issues identified automagically

docs/intro/user_types.rst
```diff
@@ -0,0 +1,123 @@
...
+
+Umm, sure. In principle, you can jump through the Basics and checkout precisely
+the sections you need, even though not all things will become clear.
+
+You want to know how to set-up and share an analysis
+with DataLad? Reading chapters :ref:`chapter_datasets`, :ref:`chapter_gitannex`,
+:ref:`chapter_yoda` and chapter :ref:`chapter_thirdparty` should work for you.
+
+You want to use DataLad as a back-up or dataset storage solution? Go to
+section :ref:`usecase_datastore`.
```
yarikoptic: cheatsheet could get someone really impatient going, I would have referenced here as well

docs/intro/user_types.rst
adswa: true, very good points.

docs/intro/user_types.rst
adswa: true, thx!

Issue #390, opened 2020-02-17
Inconvinient progress reporting for clone


adswa:
Currently, upon a re-build of the book, examples with clone look like this:

```
$ datalad clone --dataset . \
 https://github.com/datalad-datasets/longnow-podcasts.git recordings/longnow
[INFO] Cloning dataset to <Dataset path=/home/me/dl-101/DataLad-101/recordings/longnow> 
[INFO] Attempting to clone from https://github.com/datalad-datasets/longnow-podcasts.git to /home/me/dl-101/DataLad-101/recordings/longnow 
[INFO] Start enumerating objects 
[INFO] Finished enumerating objects 
[INFO] Start counting objects 
[INFO]  
[INFO]  
[INFO]  
[INFO]  
[INFO]  
[INFO]  
[INFO]  
[INFO]  
[INFO]  
[INFO]  
[INFO]  
[INFO]  
[INFO]  
[INFO]  
[INFO]  
[INFO]  
[INFO]  
[INFO]  
[INFO]  
[INFO]  
[INFO]  
[INFO]  
[INFO]  
[INFO]  
[INFO]  
[INFO]  
[INFO]  
[INFO]  
[INFO]  
[INFO]  
[INFO]  
[INFO]  
[INFO]  
[INFO]  
[INFO]  
[INFO]  
[INFO]  
[INFO]  
[INFO]  
[INFO]  
[INFO]  
[INFO]  
[INFO]  
[INFO]  
[INFO]  
[INFO]  
[INFO]  
[INFO]  
[INFO]  
[INFO]  
[INFO]  
[INFO]  
[INFO]  
[INFO]  
[INFO]  
[INFO]  
[INFO]  
[INFO]  
[INFO]  
[INFO]  
[INFO]  
[INFO]  
[INFO]  
[INFO]  
[INFO]  
[INFO]  
[INFO]  
[INFO]  
[INFO]  
[INFO]  
[INFO]  
[INFO]  
[INFO]  
[INFO]  
[INFO]  
[INFO]  
[INFO]  
[INFO]  
[INFO]  
[INFO]  
[INFO]  
[INFO]  
[INFO]  
[INFO]  
[INFO]  
[INFO]  
[INFO]  
[INFO]  
[INFO]  
[INFO]  
[INFO]  
[INFO]  
[INFO]  
[INFO]  
[INFO]  
[INFO]  
[INFO]  
[INFO]  
[INFO]  
[INFO]  
[INFO]  
[INFO]  
[...]

PR #412, opened 2020-03-11
RF publish -> push

adswa:
After https://github.com/datalad/datalad/pull/4206, we need to adjust all book examples that use `publish` to instead show `push`.
Ultimately, we want to achieve a setup that is more clever than what we have now:
- ATM, all examples that show `publish` are code-blocks. This prevents that every time the book is build from scratch, a real publish (which would also require force-pushing, and would lead to problems if anyone but me attempts it) needs to be made.
- Where we want to go to is executable code snippets and configuring the alternative Git identity with an insteadOf configuration to push to a local dataset sibling instead of the GitHub or Gin repositories that are specified. This way we can also test what breaks.

The current commit only quickly-and-dirtily replaces all references to publish with push, setting up the configuration and executable code snippets should happen in the next few days.


adswa:
Intermediate summary: it works, but given rather inconvenient output rendering (#390), its not good to merge atm.

adswa:
I think this is okay to merge for now. The first good thing: Everything still works, nothing breaks :tada:
Nevertheless, this is just the start of a refactoring, while I removed all notion of `datalad publish`, I have only replaced two instances of code blocks with actually executable code. The remaining code blocks with are pushing to special remotes (dropbox, and Github with GitLFS), but I would suggest that I tackle them when I replace/update this section with content about the rclone helper

PR #415, opened 2020-03-15
FAQ: on how to check for file content in dataset (hierarchy)

adswa:
This question was added thanks to @dorianps.

@all-contributors please add @dorianps for ideas.

There is not yet enough content like this available, but I think it can be beneficial to have - right next to the cheat sheet - a collection of complex commands that are useful for certain special cases. For now, the FAQ is the best place to it to reside, but in the long term, and "advanced" or "specialty" cheat sheet can be a more intuitive place for such code snippets.

I have added the git-annex code snippet first. In my opinion, its output is easier to parse that the json formatted report, and it returns precisely the info that is needed.


adswa:
oh man, I'm not good with bots.
Please please please @all-contributors add @dorianps for ideas 
... or at least talk to me...

PR #422, opened 2020-03-19
ENH: consistent DataLad and DataLad's

yarikoptic:
Not sure if second commit is "safe" (no side-effects due to ')


adswa:
Nice, thx! Feels safe, and I've compiled it from scratch just to check whether there is any code snippet that complains - all good. I think I will actually steal your commands and put them in a maintenance script, there are also a few other sweeps I regularly do.

PR #425, opened 2020-03-22
fix cheat sheet link and preview

aqw:
This is all untested, as I did not build the handbook (the build system scares me ;-). So all file paths are where I assume the files are.

This PR:
* normalizes the spelling of cheat sheet (all forms used were valid; but it should be consistent)
* fix the preview image for the cheat sheet (pdf is not recognized by browsers as an image type)
* simplifies the link to the PDF, so the user doesn't end up on GitHub having to click "download" to get the PDF.

This PR took the conservative approach and simplifies the process from 3 steps to 2 when acquiring the cheat sheet. If you don't have plans to add content to the cheat sheet webpage, you could  remove it entirely and just link to the cheat sheet directly, reducing the steps to 1.

### DIFF
--- a/docs/basics/101-124-procedures.rst
+++ b/docs/basics/101-124-procedures.rst
@@ -359,7 +359,7 @@ was applied.
    .. todo::
 
       It might be helpful to have (or reference) a table with all available
-      procedures and a short explanation. Maybe on the cheatsheet.
+      procedures and a short explanation. Maybe on the cheat sheet.
 
 Summing up, DataLad's :command:`run-procedure` command is a handy tool
 with useful existing procedures but much flexibility for your own
--- a/docs/basics/101-136-cheatsheet.rst
+++ b/docs/basics/101-136-cheatsheet.rst
@@ -10,5 +10,5 @@ DataLad cheat sheet
 Click on the image below to obtain a PDF version of the cheat sheet. Individual
 sections are linked to chapters or technical docs.
 
-.. figure:: ../artwork/src/datalad-cheatsheet.pdf
-   :target: https://github.com/datalad-handbook/artwork/blob/master/src/datalad-cheatsheet.pdf
+.. figure:: ../artwork/src/datalad-cheatsheet.svg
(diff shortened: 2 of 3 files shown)


adswa:
Thanks for this! Two mildly annoying things:

- We changed the figure from ``svg`` at one point because ``librsvg2-bin`` had trouble converting it. I've just tried to compile the book and luckily it works again for me atm, but both me and Michael ran into it, and it causes a spectacular build error that is really hard to figure out. Not that I expect many people to build the book*, but if they do, I wouldn't want them to be in this mess.  What's your opinion on using ``png`` or ``jpg`` as a preview?
- Even though I can build it locally, I cannot test whether the download links work (when the docs are build they live under a local ``_build/html`` directory, and relative links to other sections/this PDF are always broken in preview). My plan is to change the preview picture to ``png`` or ``jpg`` (depending on your suggestion; I'd push directly into this PR), and then merge this, check whether it works, and if not, mail a turd to your address. :-)



*exhibit A, Mr. @aqw on March 22nd 2020, and I quote directly: 

> I did not build the handbook (the build system scares me ;-).

aqw:
If doing a preview, then png is better suited for this case. If it were me, I would just kill the entire page and link to the cheat sheet directly, but I don't know what your future plans are for this page/content.

* Any turds mailed will be returned in boxes heavily scented with Sugar Beet Factory Eau de Parfum. ;-)

adswa:
Unfortunately, no relative links across submodule boundaries :woman_shrugging: Will revert the relative link, but keep the spelling fixes. Thx again!

aqw:
Then you should use `https://github.com/datalad-handbook/artwork/raw/master/src/datalad-cheatsheet.pdf` and wherever the png ended up. Otherwise you're linking to a webpage that contains the content, rather than the content itself.

adswa:
Thanks! Done in c751c0684bdef17a3e5c8d643379f3d154996617, committed with you as an author

PR #433, opened 2020-03-25
Simple helper script to create push targets

adswa:
It results in the correct output (creates a sibling, pushing works), but command execution fails (complains that sibling is already known). Need to investigate.


adswa:
Ha, I got it. Will push a fix.

adswa:
This allows to create local push targets: Sibling datasets (with annex or without, bare or non-bare) registered under an arbitrary sibling name in the dataset. After such a configuration, we can pretend to ``push`` datasets to hosting services such as github or gin, and we can test the functionality of ``push``.

Usage:

```
python3 /home/me/makepushtarget.py </path/to/dataset> <siblingname> </path/to/push/target> <annex (True/False)><bare (True/False)>
```
Example:

```
.. ifconfig:: internal

    .. runrecord:: _examples/DL-101-139-101
       :language: console

       $ python3 /home/me/makepushtarget.py '/home/me/dl-101/DataLad-101' 'gin' '/home/me/pushes/DataLad-101' True True

```

PR #445, opened 2020-04-03
New Gists

adswa:
- how to determine type of dataset
- backup strategies for datasets
- retrieving multiple files across dataset hierarchies
- parallization using unix tools


adswa:
note to self: the backup gist needs an update

adswa:
updated with paralellization with unix tools

PR #446, opened 2020-04-07
Favicon

adswa:
I *hope* that this is places the favicons in the right place and links them correctly. I have no way of previewing locally.


adswa:
Let's try.

PR #448, opened 2020-04-07
Favicon attempt 2

adswa:
Turns out, there is a configuration buried deep in sphinx docs: https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-html_favicon

Works locally now, but we'll see.


adswa:
Go!

PR #450, opened 2020-04-14
RF: Split Basics into Advanced and Basics part

### LINKED ISSUE #137: Need for an "advanced topics" part?
It occurred to me that we should decide whether or not we would consider a topic like "extending datalad" as in-scope scope for the book. It is clearly not "basic" and maybe not a use case. It would cover custom dataset procedures and writing one's own extension packages.

adswa:
Fixes #137

I have moved the chapters "Advanced options" and "Go big or go home" into a new "Advanced" part of the book. An intro details that this book part is not to be read sequentially, but rather chapter-wise, if the topics intersect with usecase or interest of the reader.
The Basics will currently be 9 chapters in total, the Advanced part will be two.

(Future) TODO:
- adjust gitignore section: strip narrative
- if accepted, move upcoming sections on hirni and ukb in advanced part
- potentially adjust chapter naming


adswa:
thx, will resolve conflicts and merge

PR #455, opened 2020-04-16
Singularity instructions 

adswa:
for INM7. Not in standard use case format, but a sensible place to put it, I guess. To be extended with whatever we learn throughout the process.


mih:
Let's keep this hanging here for now. We first need to figure out, if that is meant to be this way, and who is intended to use it like this.

adswa:
I don't think this PR is relevant for anything anymore. please reopen if you disagree, @mih

PR #461, opened 2020-04-23
Move RIA details into standalone section

### LINKED ISSUE #459: Separate RIA store section
I feel like the RIA store functionality is somewhat buried in the handbook. I think it is nicely layout in the usecase, but the location of where this functionality is described somewhat strays from the general setup of the book. I propose to move a large chunk of functionality description of RIA stores into a new section in the "Advanced options" chapter, and keep the usecase shorter, with an emphasize on high-level implementations (no detailed description of what a RIA store is, but how it is used in our usecase).

I have the feeling I also need to look in to the current state of the code and adjust the text to the RIA/ORA name change.

adswa:
Fixes #459

This PR splits the RIA datastore usecase into two parts:
- One new section on RIA stores and ORA remotes in the advanced part of the book
- A datastorage/data management usecase 

The advanced section will be dedicated to general explanations of RIA stores, technical details, and commands.
The usecase will show (on a non-technical level) how INM7 uses the RIA store concept, shed some light on our workflows, and highlight our hardware setup. 

#### TODOs
- [x] update the advanced section with all new/renamed functionality and concepts once all RIA-related DataLad PRs are merged
- [ ] update the usecase with @mih's idea of a janitor job/tool that can run on any dataset in a store and report how much data is hosted that is no longer referenced. 
> given a reference clone of a repository, I can make modifications, and can use ``git annex unused`` to  find files that are no longer needed. Following that I can do this:

> `git annex drop --unused --from inm7-storage (--force)`

> and it will drop these files from central storage the local clone doesn't need to have the file content in order to be able to do that

```

```


adswa:
@mih and @bpoldrack, this PR creates a standalone section on RIA stores. It moves all command details previously found in the datastore usecase to a section in the chapter "Further options", and now contains extensive command demonstrations. I would be grateful if you could give this new section a read: https://github.com/datalad-handbook/book/blob/2fcdd981aa3ccb4dea552cc6fc184b9c966312fb/docs/beyond_basics/101-147-riastores.rst

Please let me know if some RIA feature is missing from this section. What I'm currently aware of:
- [ ] the ``export-archive`` feature is yet undemonstrated. I hope I can get insights into this in the wake of the UKBiobank effort, and then link the chapter that comes out of this effort

The usecase has been slimmed down as a result of this. My plan is to detail workflow implementations in INM-7 instead, such as 
- demonstrations of maintenance tasks or queries (I believe there was a Gitlab repo with maintenance scripts somewhere)
- insights into custom procedures (should we still have the ``inm-7`` procedure in place)
- how standard analysis workflows look like from a user side 

However, I don't think that much of the workflow details are known at this point, yet. I hope that I can add these workflows and merge this PR prior to the ``0.13`` release.

docs/beyond_basics/101-147-riastores.rst
```diff
@@ -0,0 +1,704 @@
...
+.. note::
+
+   Setting up and interacting with RIA stores requires DataLad version ``0.13.0``
+   or higher. In order to understand this section, some knowledge on Git-internals
+   and overcoming any fear of how checksums and UUIDs look can be helpful.
+
+Technical details
+^^^^^^^^^^^^^^^^^
+
+RIA stores can be created or extended with a single command inside of any
```
bpoldrack, line 25: "from within" instead of "inside of"?
It sounds a bit as if the store would be created inside the dataset.

docs/beyond_basics/101-147-riastores.rst
```diff
@@ -0,0 +1,704 @@
...
+from standard DataLad workflows. The paragraphs below detail how to create and
+populate a RIA store, how to clone datasets and retrieve data from it, and also
+how to handle permissions or hide technicalities.
+
+Creating or publishing to RIA stores
+""""""""""""""""""""""""""""""""""""
+
+.. index:: ! datalad command; create-sibling-ria
+
+A dataset can be published into an existing or not yet existing RIA store by
```
bpoldrack, line 272: "published" is incorrect. `create-sibling-ria` doesn't publish. It only creates the sibling.

docs/beyond_basics/101-147-riastores.rst
```diff
@@ -0,0 +1,704 @@
...
+
+.. findoutmore:: What about creating RIA stores and cloning from RIA stores with different protocols
+
+   Consider setting up and populating a RIA store on a server via the ``file``
+   protocol, but cloning a dataset from that store to a local computer via
+   ``SSH`` protocol. Will this be a problem for file content retrieval?
+   No, in all standard situations, DataLad will adapt to this. Upon cloning
+   the dataset with a different protocol than it was created under,
+   enabling the special remote will initially fail, but DataLad will adaptively
+   try out other protocols to enable the ora-remote and retrieve file contents.
```
bpoldrack, line 544: "try out other protocols"

Technically that's not correct. What `datalad clone` does, is to try to enable the special remote using the same URL it cloned from. That may or may not include a protocol change. Could also be a change in hostname or path.

docs/beyond_basics/101-147-riastores.rst
```diff
@@ -0,0 +1,704 @@
...
+   The usecase :ref:`usecase_HCP_dataset`
+   details a RIA-store based publication of a large dataset, split into a nested
+   dataset hierarchy with about 4500 subdatasets in total. But how can links to
+   subdatasets work, if datasets in a RIA store are stored in a flat hierarchy,
+   with no nesting?
+
+   The key to this lies in flexibly regenerating subdataset's URLs based on their
+   ID and a path to the RIA store. The :command:`datalad get` command is
+   capable of generating RIA URLs to subdatasets on its own, if the higher level
+   dataset containts a ``datalad get`` configuration on ``subdataset-source-candidate-origin``
```
bpoldrack, line 572: "containts" has an excess `t`. ;-)

docs/beyond_basics/101-147-riastores.rst
adswa: hihi, true

docs/beyond_basics/101-147-riastores.rst
adswa: yes, thx

docs/beyond_basics/101-147-riastores.rst
adswa: thx

docs/beyond_basics/101-147-riastores.rst
adswa: ;-) thx!

PR #476, opened 2020-05-06
Add the datalad copy-file command

adswa:
This adds demonstrations of `datalad copy-file` to the section "Miscellaneous Filesystem Operations" and the HCP usecase. I think in the future there may be a better place for a standalone (sub)section on ``copy-file``, but for now these two spots in the handbook are quite suitable.


docs/usecases/HCP_dataset.rst
```diff
@@ -444,6 +449,396 @@ throughout the dataset hierarchy.
...
+      If a target directory or a destination path is given for a file, the copied
+      file will be untracked:
+
+      .. runrecord:: _examples/HCP-5
+         :language: console
+         :workdir: usecases/HCP
+
+         $ datalad copy-file hcp/HCP1200/130013/T1w/Diffusion/bvecs -t dataset-to-copy-to
+
+      Note how the file is untracked afterwards:
```
mih, line 590:
```suggestion
      If a target directory or a destination path is given for a file, the copied
      file will be not be saved:

      .. runrecord:: _examples/HCP-5
         :language: console
         :workdir: usecases/HCP

         $ datalad copy-file hcp/HCP1200/130013/T1w/Diffusion/bvecs -t dataset-to-copy-to

      Note how the file is added, but not saved afterwards:
```

PR #482, opened 2020-05-09
BF: Replace inaccurate description of push behavior

### LINKED ISSUE #481: Inaccurate description of how `push` handles multiple branches
docs/basics/101-141-push.rst (0246c38) says

```
If you have more than one :term:`branch` in your dataset, note that a
:command:`datalad push` command will by default update all branches that both
the sibling and the dataset share.
```

As far as I can tell, by looking at push.py and trying it out (e.g., the script below), that's not the case.  Putting aside the special handling of the git-annex branch, the approach is instead "try to do what `git push REMOTE` would do, falling back to pushing the current branch".  That *may* match your description, if the user configured `push.default` to `matching` (or tweaked the default push refspec), but the default for `push.default` (since Git 2.0) is `simple`, which restricts the push to the current branch.

<details>
<summary>script</summary>

```sh
cd "$(mktemp -d ${TMPDIR:-/tmp}/dl-XXXXXXX)"
datalad create a
git -C a config receive.denyCurrentBranch updateInstead

datalad clone a b
(
    cd b
    git config push.default simple  # current default value for git
    git commit --allow-empty -mc0
    git commit --allow-empty -mc1
    git checkout -b other master~1
    git commit --allow-empty -mc2
    datalad push --to origin
)
```

Output with DataLad 1b908b4ca:

```
[...]
publish(ok): . (dataset) [refs/heads/other->origin:refs/heads/other [new branch]]
publish(ok): . (dataset) [refs/heads/git-annex->origin:refs/heads/git-annex 9152532..6ecf2a2] 
```

</details>

adswa:
As described in #481, the previous description of how push
behaves in datasets with multiple branches was inaccurate,
and resulted from my explorations of push with an explicitly
configured dataset. This change rectifies this, and instead
shows how to configure datasets such that more than just the
current branch is pushed. Fixes #481.

I would appreciate if someone could double check. :)

### DIFF
--- a/docs/basics/101-141-push.rst
+++ b/docs/basics/101-141-push.rst
@@ -83,26 +83,37 @@ option. Several different modes are possible:
 - ``all``: The final mode, ``all``, combines modes ``gitpush`` and ``datatransfer``,
   thus attempting to really get your dataset contents published.
 
-.. findoutmore:: Details on push for Gitusers
+.. Findoutmore:: Pushing more than the current branch
 
-   The commands :command:`git push` and :command:`datalad push` appear similar,
-   but there are crucial differences between them. Depending on how well you
-   know Git, and how complex your DataLad workflows are, you should be aware of
-   them to pick the correct command for your use case.
+   If you have more than one :term:`branch` in your
+   dataset, a :command:`datalad push --to <sibling>` will by default only push
+   the current :term:`branch`, *unless* you provide configurations that alter
+   this default. Here are two ways in which this can be achieved:
 
-   If you have more than one :term:`branch` in your dataset, a
-   :command:`datalad push --to <sibling>` will update all branches that the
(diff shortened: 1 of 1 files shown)


kyleam:
@adswa Thanks for adjusting the text.  

Sorry I didn't provide feedback even though this has been open for a while.  While I do fetch this repo every so often and glance through all the nice additions, I'm not "watching" it, so I didn't get an email about this.  With how I intake information from GitHub, that means I didn't know this PR existed :(

adswa:
Whoops, sorry for not responding for two days already, too, @kyleam! No worries, I should have tagged you if I wanted to ask for explicit feedback. I remember thinking "It's a Saturday, and it concerns a command that isn't even released yet, so why rush/bother others". I promise that if I would appreciate feedback on something I will tag you :)

PR #485, opened 2020-05-13
Minor additions to Git LFS section on compression/encryption and drop

adswa:
Basically copy-pasted from https://github.com/datalad/datalad/issues/4489.


yarikoptic:
looks good to me, thank you!

PR #489, opened 2020-05-18
zenodo: Add Christian

adswa:
This puts him into the official, citeable author list of the book.
- [x] TODO: engage the allcontributors bot to put him onto the cover as well.


adswa:
the bot has done its work, I'll merge this as well. I've confirmed the JSON is valid, so the contributor list should be up to date and valid for the release.

PR #496, opened 2020-05-24
More book-like titlepage for the PDF

mih:
Still auto-generated, but looks less like it, I hope.

image


adswa:
wow, thats pretty. Thanks!

PR #499, opened 2020-05-25
Build two-sided PDF

mih:
First, it makes sense to targeted two-sided printing at 400+ pages, and
second it removed tons of warnings from the PDF build, because it
removes a conflict between sphinx's fancyhdr setup, and our previous
explicit 'oneside' configuration.


adswa:
okay, you convinced me. ;-)

PR #500, opened 2020-05-26
Improve PDF, no manual numbering

mih:
The point of this change is to get a cleanly structured PDF that can be navigated naturally, and does not suffer from custom (invisible) numbering. All that without impacting the HTML version too much. I think it came out OK.

image

image


adswa:
thanks a lot for this!

PR #502, opened 2020-05-28
typos to check RTD builds

adswa:
I'm checking whether RTD PR builds happen now.


mih:
Where would these RTD build end up?

adswa:
Seems like right here, as a check under each PR (based on the gif [here](https://docs.readthedocs.io/en/latest/guides/autobuild-docs-for-pull-requests.html))

adswa:
ha, nice. :)

adswa:
This feature is super cool! thx @yarikoptic!!!

mih:
Really sweet! Should we then disable the PR builds on travis? (just keeping the branch builds)?

adswa:
Intuitively I'd say no, as I suspect that Travis will result in more informative crashes and tracebacks, but let me make a commit fail and see how it looks like.

PR #503, opened 2020-05-29
Trigger a test failure with runrecord without attached example

adswa:
https://github.com/datalad-handbook/book/pull/502#issuecomment-635797007

### DIFF
--- a/docs/intro/installation.rst
+++ b/docs/intro/installation.rst
@@ -33,6 +33,12 @@ exist, and they need to be installed separately, if needed.
 .. figure:: ../artwork/src/install.svg
    :width: 70%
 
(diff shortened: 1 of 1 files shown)


adswa:
so readthedocs throws the same error as Travis does, travis just does it in color ;-) Is there any other travis benefit over readthedocs CI that I'm missing? If not, I'm indifferent about keeping travis or throwing it out, and if you like, we can rip it out @mih.

adswa:
I vote against disabling the Travis build. #532 has just shown cases in which readthedocs CI didn't catch a LaTeX-related error that made travis fail. I don't understand why exactly, but better be safe than sorry?

Issue #494, opened 2020-05-23
Permanent URLs


mih:
This is an attempt to record URLs that need to be maintained permanently (e.g. via redirects), because they have been used in unmaintainable documents (e.g. reports) -- even if the handbook needs to be restructured.

- http://handbook.datalad.org/usecases/HCP_dataset.html
- http://handbook.datalad.org/usecases/reproducible-paper.html

adswa:
I will use this issue to document a redirection feature we have just drafted.

There is a file [r.rst](https://raw.githubusercontent.com/datalad-handbook/book/master/docs/r.rst) which handles redirection.
Its an rst file with raw HTML that includes a bit of JS (I honestly know too little about this, but it feels like [this comic](https://m.xkcd.com/1926/)). 

It contains a mapping that can match a '#ID' appended to handbook.datalad.org/r.html:
```JS
        let redirects = new Map([
            ['GIN', 'basics/101-139-gin.html']
        ]);
```

opening handbook.datalad.org/r.html?GIN will redirect automatically to handbook.datalad.org/basics/101-139-gin.html with this mapping.

Mapping need to be defined with "?<id>", "relative-url-from-docs.html"

mih:
With #518 merged, we have this info in code and will keep it there.

PR #518, opened 2020-06-02
Add more structure and info to redirection page

mih:


docs/r.rst
```diff
@@ -1,24 +1,38 @@
...
-          document.getElementById("demo").innerHTML = "no idea"
-        } else {
-          window.location.replace(window.location.href.replace(/r.html\?.*/, redirect))
-        }
-        </script>
-     </head>
-   </html>
+.. raw:: html
+
+   <p id="message"></p>
```
mih, line 10: This could have a default message that anticipates a successful redirection, like "You will be redirected". Maybe we even render the target URL here to (ala, if this doesn't work, click here).

docs/r.rst
```diff
@@ -1,24 +1,38 @@
...
+   This defines a mapping of redirect codes to their present URLs
+
+.. raw:: html
+
+   <script>
+   let redirects = new Map([
+       ['GIN', 'basics/101-139-gin.html'],
+       ['HCP-dataset', 'usecases/HCP_dataset.html'],
+       ['reproducible-paper', 'usecases/reproducible-paper.html']
+   ]);
```
mih, line 22: Added what we had in #494

docs/r.rst
```diff
@@ -1,24 +1,38 @@
...
+..
+   This code replaces the r.html?key part with the final URL, while keeping
+   the rest of URL intact.
+
+.. raw:: html
+
+   <script>
+   redirect = redirects.get(window.location.href.replace(/.*\?/, ""));
+   if (redirect == undefined) {
+     document.getElementById("demo").innerHTML = "no idea"
```
mih, line 34: Message should get more informative.

docs/r.rst
adswa: yes. agreed

docs/r.rst
adswa: nice

docs/r.rst
adswa: jup

docs/r.rst
```diff
@@ -1,24 +1,71 @@
...
-   </html>
+..
+   Include a named paragraph in the page, where the javascript code below will
+   place any message.
+
+.. raw:: html
+
+   <p><strong id="successmessage">
+     You will be redirected to your target page in a few seconds.
+   </strong></p>
```
mih, line 14: we could turn this `<strong>` into CSS ...

PR #519, opened 2020-06-05
Redirects for links used in OHBM OSR talk

adswa:


mih:
I don't mind the redirect label choice. But in general I wonder if we should come up with a pattern. Some of the current choices shorten existing labels, keep them as is, use completely different ones. It all make "post-hoc sense", but maybe we want to add some guidelines in the comments of the page to prevent some randomness.

PR #532, opened 2020-06-19
Windows-Adjustments

adswa:
- adds a custom "Windows Workaround" Container (i.e., a findoutmore in blue with a different title)
- Provides windows users with a custom dataset configuration to replace `text2git`.

### DIFF
--- a/dataladhandbook_support/directives.py
+++ b/dataladhandbook_support/directives.py
@@ -129,6 +129,83 @@ class FindOutMore(BaseAdmonition):
         return [toggle]
 
 
+class WindowsWorkArounds(BaseAdmonition):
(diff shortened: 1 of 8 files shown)


adswa:
@mih could you take a look at this? I hope that the changes in this PR make the book usable for windows users again.

mih:
I think this is fine as is -- but a bit scary. I made a note in https://github.com/datalad/datalad/issues/3360 that we could simplify this, by doing exactly what you describe, but ship a procedure that readily does that.

adswa:
I'm in favor of an adjusted procedure for windows. For now I will merge this and update later if we have one.

PR #533, opened 2020-06-19
Add OHBM poster to handbook

adswa:
Its in this Repo and linked in the README, but this PR also makes it easily discoverable and downloadable from the handbook.


adswa:
I will merge this in, OHBM is already ongoing and access to the poster is helpful

PR #551, opened 2020-07-17
Mention osf extension in third party chapter

adswa:
Given how easy it is to use the OSF with the new extension, I have placed a note about the OSF into the "here are easier alternatives" box.


adswa:
Some short PR's have been open for a really long while now. Given no-one complained, I will merge them.

PR #559, opened 2020-07-31
update contributors info

Remi-Gau:
Adding myself and contribution info.

Let me know if you want to update the DOI badge to point to the "All version DOI".

### DIFF
--- a/.zenodo.json
+++ b/.zenodo.json
@@ -137,6 +137,11 @@
       "orcid": "0000-0003-0679-1985"
     },
     {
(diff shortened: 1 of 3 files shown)


adswa:
Sorry for the late reply, today was a bit unusual. :) I would appreciate an all version DOI a lot!

Remi-Gau:
> Sorry for the late reply, today was a bit unusual. :) I would appreciate an all version DOI a lot!

I am not mistaken, this should do it. :-)

PR #583, opened 2020-09-27
MNT: Update singularity image for midterm analysis to latest version

adswa:
The previous container was a year old with git-annex 7.x and ran into issues in
v8 annex repositories. This change replaces code pointing to the previous image with code pointing to the new one.

This doesn't feel like the most clean way to do things, though. Is there a way to add a ``latest`` tag to the most recent image and change the code to pull the image via this tag?


adswa:
This one is also overdue, merging.

PR #587, opened 2020-10-06
DVC: Fix-up publish-push confusion in code block

adswa:
This one slipped through, I didn't update the code example after switching from publish (used due to my "push.default = matching" configuration) back to push when we identified the problem.


adswa:
I will merge this, no one has complained.

PR #588, opened 2020-10-08
Towards more user-friendliness for Windows Users

adswa:
I'm working my way through the handbook on Windows and add "Windows-Workaround" Sections where ever necessary. So far, its frequently necessary.

### DIFF
--- a/docs/intro/installation.rst
+++ b/docs/intro/installation.rst
@@ -249,35 +249,31 @@ Expandable notes, "``Windows-Workaround``\s", contain important information, alt
 
 - **Step 2**: Install Git
 
-  - In the ``Anaconda prompt``, run::
-
-       conda install -c conda-forge git
-
-    Note: Is has to be from ``conda-forge``, the anaconda version does not
-    provide the ``cp`` command.
+  - If you haven't installed :term:`Git` yet, please download and install the latest release from `git-scm.com/ <https://git-scm.com/>`_.
 
 - **Step 3**: Install git-annex
 
-  - Obtain the current git-annex versions installer
-    `from here <https://downloads.kitenet.net/git-annex/windows/current/>`_.
-    Save the file, and double click the downloaded
-    :command:`git-annex-installer.exe` in your Downloads.
+  - Obtain the current git-annex versions installer `from here <https://downloads.kitenet.net/git-annex/windows/current/>`_.
+    Save the file, and double click the downloaded :command:`git-annex-installer.exe` in your Downloads.
 
   - During installation, you will be prompted to "Choose Install Location".
-    **Install it into the miniconda Library directory**, e.g.
-    ``C:\Users\me\Miniconda3\Library``.
+    **Install it into the directory in which Git is installed**, e.g., ``C:\Program Files (x86)\Git``.
+    Usually, :term:`git-annex` will set the correct installation location automatically, and warn if it doesn't find a Git installation where it expects one to be.
 
 - **Step 4**: Install DataLad via pip
 
   - ``pip`` was installed by ``miniconda``. In the ``Anaconda prompt``, run::
 
-       pip install datalad~=0.12
+       pip install datalad
 
-- **Step 5**: Install 7zip
+- **Step 5**: Install 7zip and Unix tools
 
-  - `7zip <https://7-zip.de/download.html>`_ is a dependency of DataLad and
-    not installed by default on Windows 10. Please make sure to download and
-    install it.
+  - `7zip <https://7-zip.de/download.html>`_ is a dependency of DataLad and not installed by default on Windows 10.
+    Please make sure to download and install it.
+    Please also make sure that it is available in your ``PATH`` environment variable: ``Environment Variables`` -> ``User Variables`` -> ``path`` -> ``edit`` -> ``new`` -> add the path: ``C:\ProgramFiles\7-zip``) -> ``OK``. Lastly, within ``C:\ProgramFiles\7-zip``, copy the ``7z.exe`` file into ``C:\Users\<user>\Miniconda3\Library\usr\bin\``.
+  - Many Unix command-line tools such as ``cp`` or ``tree`` are not available by default.
+    You can get a good set of tools by installing :term:`conda`\s ``m2-base`` package via ``conda install m2-base`` in an Anaconda prompt.
+
 
 .. _installconfig:
 
--- a/docs/intro/installation.rst
+++ b/docs/intro/installation.rst
@@ -107,7 +107,7 @@ manager ``pip`` as described below. ``pip`` should already be installed by
 default. Recent macOS versions may have ``pip3`` instead of ``pip`` -- use
 :term:`tab completion` to find out which is installed. If it is ``pip3``, run::
 
-   $ pip3 install datalad~=0.12
+   $ pip3 install datalad
 
 instead of the code snippets in the section below.
 
@@ -116,7 +116,7 @@ a user's home directory:
 
 .. code-block:: bash
 
-   $ pip3 install --user datalad~=0.12
+   $ pip3 install --user datalad
 
 
 .. findoutmore:: If something is not on PATH...
@@ -179,14 +179,14 @@ to automatically install DataLad and most of its software dependencies, type
 
 .. code-block:: bash
 
-   $ pip install datalad~=0.12
+   $ pip install datalad
 
 If this results in a ``permission denied`` error, install DataLad into
 a user's home directory:
 
 .. code-block:: bash
 
-   $ pip install --user datalad~=0.12
+   $ pip install --user datalad
 
 ``pip`` is not able to install the `7-zip <https://7-zip.org/>`_ dependency.
 Please install a flavor of 7-zip that is appropriate for your operating system (such as `p7zip <http://p7zip.sourceforge.net/>`_ for Linux or macOS) if it isn't yet installed.
--- a/docs/intro/installation.rst
+++ b/docs/intro/installation.rst
@@ -204,14 +204,16 @@ Windows 10
 """"""""""
 
 There are two ways to get DataLad on Windows 10: one is within Windows itself,
-the other is using WSL, the Windows Subsystem for Linux. We recommend the
-former, but information on how to use the WSL can be found here:
+the other is using WSL2, the Windows Subsystem for Linux, version 2.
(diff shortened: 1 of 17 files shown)


adswa:
@mih it would be nice if we could chat about installation recommendations.

adswa:
This is quite substantial, and yet far from done. I'm sure that much more can and will be done to support Windows users, but this chunk is ready to go in from my point of view. @mih, I've written the "So ... Windows... eh?" page you asked for, and it is here: https://datalad-handbook--588.org.readthedocs.build/en/588/intro/windows.html#ohnowindows ;-) Not sure whether it is what you had in mind, please check.

adswa:
In this PR (among many other things) I am switching recommendations about where to install git-annex under Windows -- instead of installing Git via conda-forge (which we recommended to get ``cd``, but installing ``m2-base`` via conda is an alternative), I have switched to a Git-for-Windows installation and recommend installing git-annex inside of it, since its more likely that users already have Git installed: https://datalad-handbook--588.org.readthedocs.build/en/588/intro/installation.html#windows-10.
If anyone has opinions on how to do the Windows installation I would appreciate hearing them. There is one issue in datalad where someone hasn't yet succeeded in installing datalad with git-annex in the miniconda library and rebooting their computer without error, and I think it would be cool if we could point them to an installation method (that we have thought about, or *even better*, experience with). Given more people have a Windows computer for testing than me (@bpoldrack, @yarikoptic, @mih?), have you made any experiences about what works and what doesn't?

I also wrote a section on overthinking installation on Windows, but I would be reluctant to merge this without no-one else having had a chance to look at it: https://datalad-handbook--588.org.readthedocs.build/en/588/intro/windows.html#ohnowindows. Feedback would be really appreciated!

yarikoptic:
> someone hasn't yet succeeded in installing datalad with git-annex in the miniconda library and rebooting their computer without error, 

What error exactly? About assistant?

adswa:
https://github.com/datalad/datalad/issues/5000

yarikoptic:
oh well, a proper long term resolution to windows installation problems would be to build git-annex in conda-forge for windows (https://github.com/conda-forge/git-annex-feedstock/issues/15), and IMHO would be the most effective investment of time (even though might take longer than writing documentation outlining how things do not work, no pun intended whatsoever). The  dances with forcing annex into anaconda environment and then changing installed files sure thing cannot be relied on just prolong the pains IMHO, even if @joeyh agrees to make git-annex become more robust in such heterogeneous installations to use system-wide (not conda's) git: https://git-annex.branchable.com/bugs/old__40__er__41___cp_from_coreutils_lacks_--reflink/#comment-9819622af7c173000af5c8337dee4b10 .

mih:
Trying this out right now. Impulse response:

Switching away from the conda install for git is an immediate monster-win for a windows noob like me. The Git installer leaves a really nicely configured system behind (TT fonts, symlinks, performance tuning) -- all stuff that I would have had to discover otherwise, and probably would have given up on long before reaching this state.

docs/intro/installation.rst
```diff
@@ -253,35 +251,31 @@ touch if you want to help.
...
 
-       pip install datalad~=0.12
+       pip install datalad
+
+- **Step 5**: Install 7zip and Unix tools
 
-- **Step 5**: Install 7zip
+  - `7zip <https://7-zip.de/download.html>`_ is a dependency of DataLad and not installed by default on Windows 10.
+    Please make sure to download and install it.
+    Please also make sure that it is available in your ``PATH`` environment variable: ``Environment Variables`` -> ``User Variables`` -> ``path`` -> ``edit`` -> ``new`` -> add the path: ``C:\ProgramFiles\7-zip``) -> ``OK``. Lastly, within ``C:\ProgramFiles\7-zip``, copy the ``7z.exe`` file into ``C:\Users\<user>\Miniconda3\Library\usr\bin\``.
```
mih, line 275: Why do I need to copy the binary? After adding the PATH everything seems to work already.

docs/intro/installation.rst
```diff
@@ -253,35 +251,31 @@ touch if you want to help.
...
+  - Many Unix command-line tools such as ``cp`` or ``tree`` are not available by default.
+    You can get a good set of tools by installing :term:`conda`\s ``m2-base`` package via ``conda install m2-base`` in an Anaconda prompt.
```
mih, line 277: In the git installer I chose to use an msys terminal by default.  The resulting environment had `tree`, but did not have `cp`. Maybe remove `tree` in the text to avoid "I can get away with this alternative approach" scenarios.

docs/intro/installation.rst
```diff
@@ -253,35 +251,31 @@ touch if you want to help.
 
 - **Step 2**: Install Git
 
-  - In the ``Anaconda prompt``, run::
-
-       conda install -c conda-forge git
-
-    Note: Is has to be from ``conda-forge``, the anaconda version does not
-    provide the ``cp`` command.
+  - If you haven't installed :term:`Git` yet, please download and install the latest release from `git-scm.com/ <https://git-scm.com/>`_.
```
mih, line 254: Just FTR: Many options are advertised by the git installer. We should leave recommendations, if we have any. Candidates are:

- true type fonts (OMG my eyes can see for the first time on windows)
- msys terminal (stellar C&P performance in comparison)
- credential manager
- FS tuning options (do they have an effect?)
- enable symlinks?

mih (commented):
First set of comments after replicating the new installation instructions in win10. Much better!!

mih:
After following the installation instructions, and running the git-annex tests I get:

```
8 out of 704 tests failed (511.91s)
(Failures above could be due to a bug in git-annex, or an incompatibility
   with utilities, such as git, installed on this system.)
```

Output is pretty garbled. OK initially (with colors), but then riddled with ASCI control sequences.

adswa:
ah shit :/ is that on the computer under your desk? I haven't actually run the git-annex test suite on my Windows computer. But this may be a very useful recommendation to give people a way to judge whether everything is okay

docs/intro/installation.rst
adswa: I trusted random people on the internet - yeah, doesn't make sense to do this, sorry. will only keep the PATH instructions.

yarikoptic:
> After following the installation instructions, and running the git-annex tests I get:
> 
> ```
> 8 out of 704 tests failed (511.91s)
> (Failures above could be due to a bug in git-annex, or an incompatibility
>    with utilities, such as git, installed on this system.)
> ```
> 
> Output is pretty garbled. OK initially (with colors), but then riddled with ASCI control sequences.

Those?
https://git-annex.branchable.com/bugs/storeKey_when_already_present_failures_on_Windows/

docs/intro/installation.rst
```diff
@@ -253,35 +251,35 @@ touch if you want to help.
...
-    `from here <https://downloads.kitenet.net/git-annex/windows/current/>`_.
-    Save the file, and double click the downloaded
-    :command:`git-annex-installer.exe` in your Downloads.
+  - Obtain the current git-annex versions installer `from here <https://downloads.kitenet.net/git-annex/windows/current/>`_.
+    Save the file, and double click the downloaded :command:`git-annex-installer.exe` in your Downloads.
 
   - During installation, you will be prompted to "Choose Install Location".
-    **Install it into the miniconda Library directory**, e.g.
-    ``C:\Users\me\Miniconda3\Library``.
+    **Install it into the directory in which Git is installed**, e.g., ``C:\Program Files (x86)\Git``.
```
yarikoptic, line 262: I went with 64bit version of git and annex so no x86

docs/intro/installation.rst
```diff
@@ -253,35 +251,35 @@ touch if you want to help.
...
+  - To verify that the build of git-annex works in your Windows system, you are encouraged to run the test suite before using git-annex on real data.
+    After installation, run ``git annex test`` in the command line.
+    There will be a lot of output; the important thing is that it should end with "All tests passed".
 
 - **Step 4**: Install DataLad via pip
 
   - ``pip`` was installed by ``miniconda``. In the ``Anaconda prompt``, run::
 
-       pip install datalad~=0.12
+       pip install datalad
```
yarikoptic, line 273: What not conda install since already using conda

docs/intro/installation.rst
yarikoptic: We need to automate, https://github.com/datalad/datalad/issues/5078

adswa:
@yarikoptic I have updated the installation instructions to "native Git (with TrueType Fonts) + miniconda + conda-forge datalad" for now. I very much support automation, but until then, shall we merge this as a temporary improvement?

yarikoptic:
sounds good for now, but you might like to add (unless you already did) a note that 
- they should use "native Git Bash" terminal to interact with DataLad, for that they should have conda be available within that shell as well (I have not backtraced my steps with starting from a clean install to see how my git bash ended up seeing conda)
- have not investigated in full yet on either there would or not be side-effect: there should be no `m2-bash` (or any other package which `posix` conda package depends on) installed ATM, due to https://github.com/ContinuumIO/anaconda-issues/issues/12124 . For that effect also, if someone enters `bash` from `m2-bash` in their conda environment, they might end up with datalad which would not see git-annex since it is installed under `...\Library\bin` and that one is  dropped by `m2-bash`. If not a note -- something FYI

adswa:
>     * they should use "native Git Bash" terminal to interact with DataLad, for that they should have conda be available within that shell as well (I have not backtraced my steps with starting from a clean install to see how my git bash ended up seeing conda)

Off, I didn't have that solution (using the Git Bash) on my radar at all (I hat Windows so much... its an endless rabbit hole of non-optimal options)! One way I found to make it work was to tick the box to add Miniconda to the PATH variable during Miniconda installation. This terminal does look nicer, but I can't get it to copy-paste, and it lacks commands like ``tree`` or ``wget``.

>     * have not investigated in full yet on either there would or not be side-effect: there should be no `m2-bash` (or any other package which `posix` conda package depends on) installed ATM, due to [ContinuumIO/anaconda-issues#12124](https://github.com/ContinuumIO/anaconda-issues/issues/12124) . For that effect also, if someone enters `bash` from `m2-bash` in their conda environment, they might end up with datalad which would not see git-annex since it is installed under `...\Library\bin` and that one is  dropped by `m2-bash`. If not a note -- something FYI

Ahhhhhh crap :facepalm:

adswa:
Ok, I will merge this now. Its definitely not done, but a huge improvement to the current status.

PR #591, opened 2020-10-19
Parallel processing of datasets

### LINKED ISSUE #472: How to `datalad run` in parallel
Having a section on how to parallelize `datalad run` calls on a cluster would be a nice addition to the handbook.

I'm currently trying to run multiple `datalad run` calls in parallel (with `--explicit` set) and getting problems with git-locks.. I have a few ideas how I could avoid those in my specific case, but I assume that this would be relevant to many people using some form of HPC.

I think the section should highlight what the potential problems are (git-locks / multiple processes trying to commit at the same time) and how to avoid them. The section on [big analyses](http://handbook.datalad.org/en/latest/beyond_basics/101-161-biganalyses.html) is related/a good template for this (e.g. it mentions the `--explicit` flag), but it does not address locking issues. 

A gist or brief outline of how to code this up would be a great start for being pointed in the right direction. 

I assume that there are "intended" programming patterns and more of "anti-patterns". Having a short do's/dont's would be also an option.

adswa:
A write up of parallel preprocessing of large datasets on computational clusters with containerized pipelines and datalad containers-run. It has an examplary walkthrough through fmriprep preprocessing on the ENKI dataset. This fixes #472.

Ping @mih.


mih (approved):
Great. Makes a lot of sense to do it in two steps! Thx

PR #602, opened 2020-11-03
BF: Fix a trailing slash in the script

adswa:
There was an error in the ENKI preproc workflow script that presumably was fixed in the most recent version of the workflow file. It made it into the handbook because I wrote this section when this fix wasn't yet introduced. @mih, I would consider this a hotfix.


adswa:
I'm currently testing this again on juseless. If this makes it work, I'll merge this later this morning.

PR #624, opened 2020-12-08
REF: add permalink for Git LFS section

adswa:
cc @yarikoptic.
This allows you to reference http://handbook.datalad.org/en/latest/basics/101-138-sharethirdparty.html#use-github-for-sharing-content with "handbook.datalad.org/r.html?LFS". Is that what you had in mind?


yarikoptic:
YES! Thank you Adina!

PR #626, opened 2020-12-09
MNT: adjust for user-passwort deprecation of GitHub

adswa:
This change comes in response to @yarikoptic's PR https://github.com/datalad/datalad/pull/5218 that removes ``create-sibling-github``'s ``--github-psswd`` argument.
It adds a note about GitHub's deprecation of User-Password authentication, mentions the removal of the option starting with ``0.14.0`` (correct?), and explains ways to store/supply the oauth access token.


yarikoptic:
most likely from 0.13.6 -- it just doesn't work at all ATM, so it is a "bug fix"

adswa:
Thx. Will merge once tests are green.

PR #637, opened 2021-01-05
Prepare 0.14

adswa:
- [x] Changelog
- [ ] Check if all contributors where acknowledged
- [ ] double-check zenodo.json validity

PR #643, opened 2021-01-29
Release 0.14

### LINKED ISSUE #642: LaTeX fails
It started to surface in #637: the rule "make latexpdf" fails. Locally, I can reproduce this, with LaTeX running out of memory (my initial blame is targeted at 13 thousand push messages in the DVC usecase). I have only briefly engaged with it, and have only tried to switch to the lualatex PDF engine, but this sadly isn't of help.

adswa:
This PR supersedes #637. Hopefully, it resolved the Latex error, and thus fixes #642.


adswa:
Great, this looks good to go.

PR #647, opened 2021-02-07
Minor changes re code-blocks in PDF

mih:
A picture says more than 1000 words:

image

image


adswa:
thx. looks indeed nicer this way.

Issue #649, opened 2021-02-09
Pointers on how to install python


Remi-Gau:
Got a few labmates who have started to play around with datalad (I promise I did not twist their arms... that much).

But several of them had only the "basic" python install on their machine.

Made me realize that there is no section or pointer in the handbook to how to install python.

I think that at least a link to something, that tells people how to install python on all OS could a good thing.

I thought of the "install" section (see very bottom of [this page](https://ohbm.github.io/hackathon2020/logistics/)) we used at the last OHBM hackathon (actually reused from another event in Montreal). It focuses on using conda.

It is a bit lost in the whole page there so if you want I could try to condense it into a single paragraph for the handbook.

Happy to add pointers to any better resource you might know though.

adswa:
> Made me realize that there is no section or pointer in the handbook to how to install python.

That is very true! Thanks! We haven't actually ever thought about this outside of the context of Windows (#638). 

> I thought of the "install" section (see very bottom of this page) we used at the last OHBM hackathon (actually reused from another event in Montreal). It focuses on using conda.

Focusing on conda is a good idea. The handbook already does it for Windows and for Servers. 
Thinking about the most useful and maintainable way to add it to the installation section, I'd suggest a single "Python" paragraph at the very start. It could

- note that Python3 is required
- note that this could be checked by typing "python" and/or into the terminal
- point to Miniconda as the way to go if Python isn't installed, or only available as Python2.

I'm in the source code at the moment anyway, will try my hands at a PR :)

Remi-Gau:
Sheeesh !!! You're quick!!!!

PR #651, opened 2021-02-09
Python installation instructions

adswa:
Would you mind reviewing this, @Remi-Gau? I tried to add installation instructions for Python 3 based on your pointers and thoughts in #649. :)

### DIFF
--- a/docs/intro/installation.rst
+++ b/docs/intro/installation.rst
@@ -48,7 +48,7 @@ The easiest way to do this is to open a terminal and type "python" to open a min
 If this fails, or reports a Python version with a leading ``2``, such as ``Python 2.7.18``, try typing ``python3`` (which some systems use to disambiguate between Python 2 and Python 3).
 If this fails, too, you need to obtain a recent release of Python 3.
 
-If you are on Windows, please note that you should **not** install Python via the Windows store, even if it opens after you typed ``python`` as this version requires additional configurations by hand (in particular of your ``$PATH`` :term:`environment variable`.
+If you are on Windows, please note that you should **not** install Python via the Windows store, even if it opens after you typed ``python``, as this version requires additional configurations by hand (in particular of your ``$PATH`` :term:`environment variable`).
 Please instead check the Windows section at the end of the page for more convenient alternatives.
 
 .. findoutmore:: Python 2, Python 3, what's the difference?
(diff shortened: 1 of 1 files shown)


Remi-Gau (approved):
This looks good to me. Asked some of my lab to cross check (if they have the time).

One thing that could be worth mentioning is that it is not because they have installed Python 3 that people can / should uninstall Python 2 (unless they want to break things on their system that still relies on Python 2). I have been asked that question a couple of times already, so I suspect others will ask this question too. What do you think?

adswa:
> One thing that could be worth mentioning is that it is not because they have installed Python 3 that people can / should uninstall Python 2 (unless they want to break things on their system that still relies on Python 2). I have been asked that question a couple of times already, so I suspect others will ask this question too. What do you think?

I think it makes a lot of sense mentioning this! Will add this, thanks :)

adswa:
@all-contributors please add @Remi-Gau for review

docs/intro/installation.rst
```diff
@@ -32,6 +33,48 @@ exist, and they need to be installed separately, if needed.
...
+   Python 2 is an outdated, in technical terms "deprecated", version of Python.
+   Although it still exist as the default Python version on many systems, it is no longer maintained since 2020, and thus, most software has dropped support for Python 2.
+   If you only run Python 2 on your system, most Python software, including DataLad, will be incompatible, and hence unusable, resulting in errors during installation and execution.
+
+   But does that mean that you should uninstall Python 2?
+   **No**.
+   Keep it installed, especially if you are using Linux or MacOS.
+   Python 2 existed for 20 years and numerous software has been written in it.
+   It is quite likely that some basic operating system components or legacy software on your computer is depending on it, and uninstalling a preinstalled Python 2 from your system will likely render it unusable.
+   Install Python 3, and have both versions coexist peacefully.
```
Remi-Gau, line 65:
```suggestion
   Install Python 3, and have both versions coexist peacefully.
```

hahaha 

Did you mean to write:

> ... coexist peacefully (as long as you use some sort of virtual environment, otherwise things could get a lot less peaceful very quickly). 

:wink: 

Just kidding. This is perfect.

Remi-Gau (approved):
Good to merge for me. 

Thanks @adswa  !!!

adswa:
Great, thx!

PR #655, opened 2021-02-15
More fancy boxes in PDF

mih:
- Instead of always the same title, use an actual box title
- Annotate each box with a type icon (not just color coding for people
  that do not or cannot pay attention to color)
- Give every gitusernote a title
- RF Latex code to make it less verbose
- Include tiny logos directly
  - https://commons.wikimedia.org/wiki/File:Windows_logo_-_2012_derivative.svg
    (public domain)
  - https://www.svgrepo.com/svg/65456/idea (CC0)
  - https://git-scm.com/downloads/logos (CC-BY-SA)

image


adswa:
This looks so cool! THANK YOU!

PR #656, opened 2021-02-16
Further streamlining boxes

mih:
- enable future display of box listings (by type)

image

Note that such a listing is not actually generated in this PR, only the prerequisites are established. A related, but unaddressed, topic is the numbering of boxes and their in-text referencing. This would be most helpful in the PDF to make box positioning more flexible.


adswa:
Awesome!
mental note for myself: https://github.com/rhopfer/sphinx-numbered-blocks might be helpful for in-text references

PR #657, opened 2021-02-16
Numbered, listable and referencable boxes

mih:
image

image

image

Works in HTML and Latex.

The reference a box, simply give it a name and use that one:

```rst
  .. windowsworkarounds:: Funky instructions
     :name: funky

can be referenced with :ref:`some in-text description <funky>`
```


adswa:
The PDF looks better than the webversion now... :)

PR #658, opened 2021-02-17
Optional float mode for PDF boxes

mih:
By default all boxes are included at the location the appear in the
text. Optionally, (large) boxes be turned into float, like this:

```rst
.. windowsworkarounds:: For Windows users only
   :float: p
```

where `p` is an (optional) positioning argument (the usual LaTex kind,
i.e. could also be something like `tbp`). If the position label is not
given (i.e. just `:float:`) LaTex will do its default thing.

Floating boxes likely will need to be referenced in the text, hence
should also come with a `name` attribute:

```rst
.. windowsworkarounds:: For Windows users only
   :name: funky
   :float: p
```

This change has no implications for the HTML output.

### DIFF
--- a/dataladhandbook_support/directives.py
+++ b/dataladhandbook_support/directives.py
@@ -2,6 +2,32 @@
 
 from docutils import nodes
 from docutils.parsers.rst.directives.admonitions import BaseAdmonition
+from docutils.parsers.rst.directives import unchanged
+
+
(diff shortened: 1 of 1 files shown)


adswa:
That's cool! I will transform very long findoutmore's or those with less relevance to the text into floats.

PR #659, opened 2021-02-17
No vertical lines in a table

mih:
This is a first demo of how it could be done. Essentially given sphinx a dedicated clue via `tabularcolumns` instead of having it guess.

https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-tabularcolumns

It is using the sphinx-specifc `\Y` column width specifier that should work across all table type we need (sphinx is switching type a lot, depending on table content).

### Demo

Before:

image

Now:

image

More general table beautifications are coming, but this kind must be done per each table.


adswa:
oh, wow, that already works wonders!

mih:
Death to vertical lines!

PR #660, opened 2021-02-17
Bettertables

mih:
- de-uglify all tables
- bold table headers
- all tables referenced in-text
- table/figure number per-chapter (like everything else)
- list all boxes, figures, tables in appendix


adswa:
thank you!

PR #662, opened 2021-02-18
New boxtype `importantnote`

mih:
To replace the regular `note` directive, with support for all features
of the other boxtypes.

```rst
.. importantnote:: <title>

   Content....
```

image


adswa:
so much nicer :heart_eyes:

adswa:
I just got around to try it now. It looks nice in the PDF, but in HTML, it is a quite bleak, expandable box that hides its contents rather than making aware of something important: 
Screenshot from 2021-02-19 14-29-46

mih:
OK, there you go! That is the website now, not (just) the PDF ;-)

image

image

adswa:
Yay! :) I'll wait for the tests to finish, but its LGTM from my point of view. Really really beautiful :)

PR #664, opened 2021-02-20
note -> importantnote

mih:
Except for a few complicated cases


adswa:
The complicated cases should have been taken care of by #666

PR #671, opened 2021-02-23
Prevent multiple definition of useless labels

mih:


mih:
I will merge this now, it should have no effect on anything other than warnings.

PR #689, opened 2021-03-15
Remove quotes from config file displays

### LINKED ISSUE #688: Confusing depiction of quotes for procedure configs
As reported in https://github.com/datalad/datalad/issues/5493, the "Find out more" section "Write your own procedures" in https://handbook.datalad.org/en/latest/basics/101-124-procedures.html, somewhat suggests that within the respective config file quotes are needed for the help string and call format string of a procedure. This seems to be suggested by this part:

> Below is a minimal .datalad/config entry for a custom procedure:

> `[datalad "procedures.<NAME>"]`
> `     help = "This is a string to describe what the procedure does"`

However, neither are those quotes required in the config file, nor would the later shown call to `git-config`

> `$ git config -f .datalad/config datalad.procedures.example.help "A toy example"`

result in such quotes, nor can the quotes be achieved by using `ConfigManager.set/add` via datalad's python API.
Quotes are only needed within such command line calls (to not make "A toy example" 3 arguments to `git-config`), not within the config files itself. They don't hurt when put there by editing the config file either - they simply make no difference when we read the config.

ping @pvavra

bpoldrack:
Quotes encompassing values in git config files (including
.datalad/config), are not required and in fact inconsequential when
being read again via git-config. Showing them when displaying such
configs in the handbook may therefore be confusing and leading to the
user trying to achieve this, which is not possible via git-config or
datalad's python API. This would only be achievable by editing the raw
config file w/o a sanitizing interface.

Simply remove the depiction of such quotes, as I think debating this
aspect only makes the impression that it is somehow important, when
really these quotes have no effect at all.

(Closes #688)

### DIFF
--- a/docs/basics/101-124-procedures.rst
+++ b/docs/basics/101-124-procedures.rst
@@ -214,7 +214,7 @@ was applied.
      .. code-block:: bash
 
         [datalad "procedures.<NAME>"]
-           help = "This is a string to describe what the procedure does"
+           help = This is a string to describe what the procedure does
 
    - By default, on GNU/Linux systems, DataLad will search for system-wide procedures
      (i.e., procedures on the *system* level) in ``/etc/xdg/datalad/procedures``,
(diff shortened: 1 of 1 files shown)


adswa (approved):
Thanks a lot!

PR #694, opened 2021-03-24
Make the index more comprehensive

mih:
I am going through the book, annotating different pieces (like individual commands) with index markup, such that content in the sections with narrative become more accessible to readers that do not (intent to) consume the whole book.

This PR could grow big. But there is not need to make it anyhow complete first. It could be merged (repeatedly) whenever it makes sense.


mih:
I am not sure when I can get to this next. Let's merge it now, and not be a source of future conflicts.

PR #717, opened 2021-05-20
ENH: Use Zenodo record that resolves to latest version

effigies:
https://zenodo.org/record/3512712 resolves to 0.12.0rc6. Using record ID noted at the bottom here:
datalad_versions

I assumed that you explicitly preferred the Zenodo URL over DOI, so updated the record number without switching to DOI. I'm not sure about dropping the explicit link syntax... RST should render this as a link, but possibly there's a good reason for keeping it as an explicit RST link?

### DIFF
--- a/docs/basics/101-180-FAQ.rst
+++ b/docs/basics/101-180-FAQ.rst
@@ -156,7 +156,7 @@ How can I cite DataLad?
 
 There is no official paper on DataLad (yet). To cite it, please use the latest
 `zenodo <https://zenodo.org>`_ entry found here:
-`https://zenodo.org/record/3512712 <https://zenodo.org/record/3512712>`_.
+https://zenodo.org/record/808846.
 
 .. _dataset_textblock:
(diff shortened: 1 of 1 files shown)


effigies:
Seems to render fine in HTML: https://datalad-handbook--717.org.readthedocs.build/en/717/basics/101-180-FAQ.html?highlight=cite#how-can-i-cite-datalad

Can't see a PDF rendering to check that the URL still renders there...

adswa:
Ha, oh man, I never realized this. Thanks! Hopefully, this FAQ is resolved soon (the JOSS review started! https://github.com/openjournals/joss-reviews/issues/3262)

> Can't see a PDF rendering to check that the URL still renders there...

It is :)

effigies:
Looks like the build needs to be bumped: https://readthedocs.org/projects/datalad-handbook/builds/

adswa:
thx!

PR #718, opened 2021-05-20
ENH: Relative links in sidebar

effigies:
Found some absolute links that made it hard to find the FAQ in the rendering of #717. Feel free to close without discussion if there's a good reason to use absolute links.


adswa (approved):
>  if there's a good reason to use absolute links.

There is not good reason whatsoever - Thank you!

PR #740, opened 2021-07-21
Fix errors in 9.3 back and forth in time

AKSoo:
* git reset: Removes commits _since_ not _until_ COMMIT
* commit range wording


adswa:
Thanks so much for these fixes! :pray:

adswa (approved):

PR #747, opened 2021-08-11
Update 101-115-symlinks.rst

lilikapa:
Identified a typo in line 227: There was a double "about" before hashes ;-)


adswa (approved):
Thanks! :-)

PR #749, opened 2021-08-23
"ultimatively" => "ultimately"

arokem:
I believe this is the correct word here.


adswa (approved):
Thanks for catching this!

PR #595, opened 2020-10-21
Minor typo and other feedback on the INM7 version

jadecci:
Fix issue #593 

- Fix 2 typos
- Add a comment in Chapter 3.2 for md5 checksum in macOS
- Add a note in Chapter 7.2 to remind user to install `datalad-container` first

PR #751, opened 2021-08-24
This is the name of the file that was created earlier on.

arokem:
Looks like this was also addressed in https://github.com/datalad-handbook/book/pull/595, but seems like that one never got merged into master (but instead went to the inm7 branch).


adswa (approved):
thanks for catching this!

Issue #706, opened 2021-05-03
Use GIN as autoenabled data source


mih:
Something like this would work:

```
datalad siblings add \
  -d . \
  --name gin-update \
  --pushurl git@gin.g-node.org:/studyforrest/aggregate-fmri-timeseries.git \
  --url https://gin.g-node.org/studyforrest/aggregate-fmri-timeseries.git \
  --as-common-datasrc gin
```

But ATTN: https://github.com/datalad/datalad/issues/5621

Also, when there the GIN project is still empty (not initialized with any default content), git-annex does not find an annex and configures it to be ignored (annex-ignore). This prevents `datalad push` from uploading any files.

So:

`git config --unset-all remote.gin-update.annex-ignore`

adswa:
I have PRed #707

PR #753, opened 2021-09-01
FIX: GIN siblings - fix typo in code and note that ``--url`` should not end with ``.git``

oesteban:
Fixes these minor issues found in the handbook.


adswa:
Thanks for catching this, @oesteban! 

@mih, I remember this PR came from  #706 during the Studyforrest Hackathon, when we created the Gin repos as an alternate datasource in case Juelich's datapub server is offline again at some point. I haven't checked all the studyforrest repos, but I can't find a dataset with a configured special remote to Gin (did we forget to do this somehow?). We maybe want to check if there are datasets configured like this. And if we forgot to configure the datasets, we should probably finally get to it :-) 

@all-contributors please add @oesteban for bug .
@oesteban, please feel free to [add yourself to the zenodo.json file in the repository](http://handbook.datalad.org/en/latest/contributing.html#acknowledging-contributors).

adswa (approved):

PR #759, opened 2021-09-10
Numbered boxes and references

mih:
Now also using prefixes per box-type. Here is how it looks

image


adswa:
oh cool, thanks a lot!

PR #785, opened 2021-11-23
Add a note on how to currently glob for several levels of directories

adswa:
A quick PR to at least somewhere in the handbook make a note about the current, potentially confusing behavior in ``run``s globbing. Thanks to @AKSoo for raising it. Its probably still quite hidden and easy to miss, but I guess better than nowhere.


AKSoo:
This definitely helps for current users.

Will need further edits when recursive `**` is implemented in a future release.

adswa:
@all-contributors please add @AKSoo  for review

Issue #779, opened 2021-11-15
Problem replicating the "Using Gin as a data source behind the scenes" walkthrough


mslw:
### Current behavior

I tried to follow the [Using Gin as a data source behind the scenes](https://handbook.datalad.org/en/latest/basics/101-139-gin.html#using-gin-as-a-data-source-behind-the-scenes) walkthrough from Chapter 8.6 (Dataset hosting on GIN) verbatim, but didn't succeed. There is an error when configuring the gin remote as a common data source, which I don't know how to correct. If I ignore it and carry on, the dataset gets pushed to GitHub & Gin, but without the link between the two (i.e. getting from GitHub clone doesn't work). Commands and their outputs below.

As a side note - I assume that the git + gin part of the walkthrough is meant to be stand-alone (starts by creating a new repository on Gin & uses different repo names than the examples before). If, however, the steps depends on some previous configuration, then the wording might need tweaking.

```
datalad siblings add -d . --name gin-update --pushurl git@gin.g-node.org:/msz/gin-as-data-source.git --url https://gin.g-node.org/msz/gin-as-data-source --as-common-datasrc gin
[INFO   ] Could not enable annex remote gin-update. This is expected if gin-update is a pure Git remote, or happens if it is not accessible. 
CommandError: 'git -c diff.ignoreSubmodules=none annex initremote gin type=git location=https://gin.g-node.org/msz/gin-as-data-source autoenable=true -c annex.dotfiles=true' failed with exitcode 1 under /home/mszczepanik/Documents/tmp/gg
initremote gin 
failed
git-annex: could not find existing git remote with specified location
initremote: 1 failed

datalad siblings                                                                                                                      1 !
.: here(+) [git]
[WARNING] Could not detect whether gin-update carries an annex. If gin-update is a pure Git remote, this is expected.  
.: gin-update(-) [https://gin.g-node.org/msz/gin-as-data-source (git)]

git config --unset-all remote.gin-update.annex-ignore

datalad push --to gin-update
copy(ok): hubert-neufeld-j-udI4zim2E-unsplash.jpg (file) [to gin-update...]                                                                 
copy(ok): jay-ruzesky-9zTafGVsv-c-unsplash.jpg (file) [to gin-update...]                                                                    
copy(ok): paul-carroll-Y-nyDv3TWm0-unsplash.jpg (file) [to gin-update...]                                                                   
[...]

PR #793, opened 2021-12-08
Update Gin as data source workflow

mslw:
Suggested add - push - configure - push as a correct sequence. Addresses part of #779 and datalad/datalad#6269

I verified the proposed workflow with DataLad version 0.15.3.

Now the question is what to do about upcoming changes in DataLad 0.16. The new release introduces `create-sibling-gin` and shortens the workflow in question (`git config --unset-all remote.gin.annex-ignore` should be unnecessary when `create-sibling-gin` is used, see datalad/datalad#6230). I am in favour of keeping the current workflow (where the Gin repo is created manually) even after the release (as manual creation is not going away). Then, either add a separate example starting with `create-sibling-gin`, or note that the `git config` step can be skipped if the automated method was used. Should it be done in advance, or after DataLad 0.16 release?

### DIFF
--- a/docs/basics/101-139-gin.rst
+++ b/docs/basics/101-139-gin.rst
@@ -261,20 +261,27 @@ You will need to have a Gin account and SSH key setup, so please take a look at 
 Then, follow these steps:
 
 - First, create a new repository on Gin (see step by step instructions above).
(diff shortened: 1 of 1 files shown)


adswa (approved):
thanks a lot!

adswa:
> I am in favour of keeping the current workflow (where the Gin repo is created manually) even after the release (as manual creation is not going away). Then, either add a separate example starting with `create-sibling-gin`, or note that the `git config` step can be skipped if the automated method was used. Should it be done in advance, or after DataLad 0.16 release?

I'm also in favor of leaving it as it is with your PR. Usually, it takes more time than anticipated for the release to happen, and we can fix or add to out of date instructions once they actually got out of date :+1:

adswa:
@all-contributors please add @mslw for content

Issue #821, opened 2022-04-07
Build failed, probable dependency version conflict


mslw:
The latest [monthly link check](https://github.com/datalad-handbook/book/actions/runs/2077281692) failed with:

```
ImportError: cannot import name 'environmentfilter' from 'jinja2' (/opt/hostedtoolcache/Python/3.10.2/x64/lib/python3.10/site-packages/jinja2/__init__.py)
```

Google leads to [this issue in Sphinx](https://github.com/sphinx-doc/sphinx/issues/10291) which says that deprecations in Jinja2 are the cause. Suggested workaround is to pin `jinja2 < 3.1` (consistent with the fact that the failed build used 3.1.1, while the last successful used 3.0.3). A proper fix would be to unpin Sphinx itself (currently we have `sphinx==2.1.2`) and use the latest version, currently 4.5 (which is a big leap).

If readthedocs build fail too, we will need to pin Jinja2, and maybe consider updating Sphinx later on.

mslw:
Yes, this repeated for handbook build on github actions after a recent PR. Read the docs builds seems unaffected.

The corresponding RTD build succeeded and the handbook has recent changes. Could be because on RTD we use python 3.7 (specified in `.readthedocs.yaml`) and that affects installed libraries, but I'm not sure.

PR #827, opened 2022-04-13
Remove version pinning of Sphinx

### LINKED ISSUE #822: Consider updating Sphinx requirement
The Sphinx version used by the book is pinned to 2.1.2, while the latest Sphinx version is 4.5. According to commit history, the version was pinned due to #379. However, it might have now led to #821 (in short: we pin Sphinx, but not Jinja2 versions, which led to a mismatch).

Regarding #821, pinning Jinja is a much simpler solution - and, considering that it's *probably* a dependency of Sphinx, and not ours, pinning them both actually makes sense.

Regardless, between Sphinx 2.1.2 and 4.5.0 there were many incompatible changes, some of which might be breaking for us, so an update may be challenging.

But I'm wondering what makes more sense in a long term (staying on Sphinx 2 vs upgrading to 4), and what should an upgrade process look like (build on 4, see if errors appear, and work from there?).

I don't necessarily think we need to update (and certainly not in a hurry), but I'm leaving this thought out there.

adswa:
in an attempt to fix #822, #821

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -532,5 +532,5 @@ plantuml_latex_output_format = 'pdf'
 
 
 def setup(app):
-    app.add_stylesheet('custom.css')
+    app.add_css_file('custom.css')
     app.add_config_value('internal', '', 'env')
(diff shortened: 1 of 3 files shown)


adswa:
Nice, I think that worked :)

Issue #833, opened 2022-04-14
Document installation instructions for datalad on M1 macs


jsheunis:
Relevant initial issue in datalad repo: https://github.com/datalad/datalad/issues/5701

Chat transcript of datalad office hour discussion to help a user install datalad on an M1 mac:

---

Me to Everyone (16:02)
Sorry my connection is crappy
I heard my name
Will try move to better connection
Alex Waite to Everyone (16:06)
The preferred way to use Rosetta at the CLI seems to be: https://osxdaily.com/2020/11/18/how-run-homebrew-x86-terminal-apple-silicon-mac/
You basically alias "Terminal" and tell macOS to launch it using Rosetta.
So it's spawned in that env, so all commands will detect the arch as x86.
And then you install homebrew, conda, etc in it.
If rosetta isn't yet installed, you can run the following to install it:
/usr/sbin/softwareupdate –install-rosetta –agree-to-license
Alex Waite to Everyone (16:15)
You can test whether you're in rosetta mode or note by running:
sysctl -n sysctl.proc_translated
(1 is rosetta, 0 is not)
Michael Hanke to Everyone (16:21)
https://wwwimage-secure.cbsstatic.com/thumbnails/photos/w1920/marquee/1033010/template_sp_hero_landscape.jpg
Alex Waite to Everyone (16:22)
:-)
Michael Hanke to Everyone (16:25)
https://anaconda.org/conda-forge/datalad
Alex Waite to Everyone (16:25)
-vvv
conda search datalad
Niels R to Everyone (16:32)
what's the conda version?
Adina Wagner to Everyone (16:33)
conda config --set --env subdir=osx-64
Niels R to Everyone (16:35)
--set subdir osx-64 --env?
Michael Hanke to Everyone (16:35)
https://stackoverflow.com/questions/65415996/how-to-specify-the-architecture-or-platform-for-a-new-conda-environment-apple
Me to Everyone (16:37)
conda config --env --set subdir osx-arm64
conda config --env --set subdir osx-64
Michael Hanke to Everyone (16:42)
conda install -c conda-forge git-annex
https://anaconda.org/conda-forge/git-annex (only linux)
Alex Waite to Everyone (16:44)
which brew
Michael Hanke to Everyone (16:45)
https://stackoverflow.com/questions/64882584/how-to-run-the-homebrew-installer-under-rosetta-2-on-m1-macbook
Me to Everyone (16:45)
arch -x86_64 brew install git-annex
[...]

PR #834, opened 2022-04-15
Installation: Detail the current workarounds for installations on M1 Macs

adswa:
This arose from #833, which revealed an urgent need to document the difficulties of installing DataLad on M1 macs due to the inavailability of git-annex from brew. Thanks to  @anadiasmaile's patience, we walked through the installation instructions for a standalone git-annex install created by Fede and Vera in the Office hour, and a new paragraph on M1 macs links to those instructions now.

### DIFF
--- a/docs/intro/installation.rst
+++ b/docs/intro/installation.rst
@@ -188,9 +188,10 @@ Windows users, please `get in touch
 warnings. If you on a native Windows 10 system, you should pay close
 attention to them.
 
+.. _mac:
 
-Mac OSX
-"""""""
+Mac OSX (non-M1)
+""""""""""""""""
 
 Modern Macs come with a compatible Python 3 version installed by default. The
 :find-out-more:`on Python versions <fom-py2v3>` has instructions on how to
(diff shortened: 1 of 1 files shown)


docs/intro/installation.rst
```diff
@@ -269,6 +270,15 @@ Python's package manager <fom-macosx-pip>`.
...
            /Users/MYUSERNAME/Library/Python/3.7/bin
 
+M1 Mac OSX
+""""""""""
+
+As of April 2022, :term:`git-annex` isn't yet available from the `homebrew <https://brew.sh>`_ package manager for M1 Macs, complicating the installation process.
+Our current recommendations are:
+
+#. Check `formulae.brew.sh/formula/git-annex#default <https://formulae.brew.sh/formula/git-annex#default>`_ if there is bottle installation support provided for Apple Silicon. If there is, rely on the installation instructions for :ref:`mac` instead, and consider `opening an issue <https://github.com/datalad-handbook/book/issues/new>`_ to let us know about this.
+#. Perform a manual installation of git-annex. `This step-by-step instruction <https://github.com/fraimondo/csguide-hands-on#problem-3--git-annex-on-m1-mac>`_ has been successfully used to obtain fully functional git-annex installations (with thanks to Fede Raimondo and Vera Komeyer).
```
jsheunis, line 280: Should we (either now or in future) have these instructions physically in the book as well, as a subsection of http://handbook.datalad.org/en/latest/intro/installation.html#mac-osx?

jsheunis: i.e. not linked, in case Fede/Vera's resource changes

adswa: My gut reaction is no, because we don't even have the infrastructure to ensure they still work - but not a super strong opinion

jsheunis: I didn't think of that. So: temporary until proven to work.

yarikoptic: I gave it a try.  I think the main point from https://github.com/fraimondo/csguide-hands-on#problem-3--git-annex-on-m1-mac is the

```
/usr/sbin/softwareupdate --install-rosetta --agree-to-license
```
and here is some info on Rosetta :

Rosetta is a dynamic binary translator developed by Apple Inc. for macOS, an application compatibility layer between different instruction set architectures. [Wikipedia](https://en.wikipedia.org/wiki/Rosetta_(software))

After that I used datalad-installer

```shell
python3 src/datalad_installer.py git-annex -m autobuild
```
from its github clone (didn't have pip handy), adjusted PATH to point to that installation (didn't want to mess with /usr/local), installed datalad via conda, and `datalad search haxby` worked out as it must ;)

yarikoptic: filed issue in datalad-installer to make it turnkey  https://github.com/datalad/datalad-installer/issues/103

adswa:
Merging for now - any improvements welcome via PR!

PR #836, opened 2022-04-20
minor fixes to 'Amazon S3 as a special remote'

AKSoo:
add `=`
remove `<>` (may confuse non-technical users)


adswa (approved):
That makes sense, thanks a lot!

PR #846, opened 2022-06-19
git-annex installation instructions for M1 Mac

Arshitha:
Adding a link to an additional step-by-step instruction guide to install git-annex on M1 Mac using Homebrew and Rosetta 2 via command line

PR #848, opened 2022-06-29
Update .zenodo.json

Arshitha:
@adswa sorry for the new PR! I thought I'd made the change on the same branch as the [old PR](https://github.com/datalad-handbook/book/pull/846) but apparently not. Would this work?

### DIFF
--- a/.zenodo.json
+++ b/.zenodo.json
@@ -202,14 +202,14 @@
       "orcid": "0000-0001-7041-598X"
     },
     {
-      "name": "Hanke, Michael",
-      "affiliation": "Institute of Neuroscience and Medicine, Brain & Behaviour (INM-7), Research Centre Jülich, Jülich, Germany and Institute of Systems Neuroscience, Medical Faculty, Heinrich Heine University Düsseldorf, Düsseldorf, Germany",
-      "orcid": "0000-0001-6398-6370"
-    },
-    {
       "name": "Basavaraj, Arshitha",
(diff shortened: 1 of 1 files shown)


adswa:
No need to be sorry for a new PR at all - instead, thanks for taking the time to propose it! I only have one small remark. Can you move your entry to the second-to-last (because we keep last and first author fixed)? (That's also my bad, sorry, I should have specified this precisely in my comment)

.zenodo.json
```diff
@@ -201,16 +201,16 @@
...
-    {
-      "name": "Hanke, Michael",
-      "affiliation": "Institute of Neuroscience and Medicine, Brain & Behaviour (INM-7), Research Centre Jülich, Jülich, Germany and Institute of Systems Neuroscience, Medical Faculty, Heinrich Heine University Düsseldorf, Düsseldorf, Germany",
-      "orcid": "0000-0001-6398-6370"
-    },
     {
       "name": "Basavaraj, Arshitha",
       "affiliation": "Data Science and Sharing Team, National Institute of Mental Health, National Institutes of Health, Bethesda, MD, USA.",
       "orcid": "0000-0002-6984-7969"
     }
```
adswa, line 208: Just one missing comma:
```suggestion
    },
```

Arshitha: Thanks @adswa ! I generally just do `Commit Suggestion` but I appreciate you mentioning the equivalence between new commit and push from the fork. It's easy to forget that with one-click solutions. :)

adswa (approved):

PR #855, opened 2022-07-31
Mention alias parameter of create-sibling-ria

### LINKED ISSUE #845: Mention alias option in the `create-sibling-ria` section
### Is your feature request related to a problem?

The cloning-and-updating-from-ria-stores section as `find out more` about setting up aliases

http://handbook.datalad.org/en/latest/beyond_basics/101-147-riastores.html#cloning-and-updating-from-ria-stores

But it does not mention the alias option of the `datalad create-sibling-ria`

```
datalad create-sibling-ria --help 
 
  --alias ALIAS         Alias for the dataset in the RIA store. Add the necessary symlink so that this dataset can be cloned from the 
                                RIA store using the given ALIAS instead of its ID. With `recursive=True`, only the top dataset will be aliased.
                                Constraints: value must be a string or  value must be NONE
```

### Describe the solution you'd like

If those refer to the same alias, this should probably be mentioned.

I like that the find out more explains what is happening under the hood, ensuring that the alias option actually implements it this way.

### Describe alternatives you've considered

_No response_

### Additional context

_No response_

adswa:
Fixes #845

### DIFF
--- a/docs/beyond_basics/101-147-riastores.rst
+++ b/docs/beyond_basics/101-147-riastores.rst
@@ -307,6 +307,9 @@ on where the RIA store (should) exists, or rather, which file transfer protocol
 
 Note that it is always required to specify an :term:`absolute path` in the URL!
 
+In addition, as a convenience for cloning, you can supply an ``--alias`` parameter
+with a name under which the dataset can later be cloned from the dataset.
+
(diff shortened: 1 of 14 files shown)


Remi-Gau:
Thanks for that update. :hugs:

PR #878, opened 2022-12-06
Section 1.5 patch

candleindark:
I have corrected a highlight and replaced some wordings. Again, the rewordings are subjective. Take them only if you see fit.


adswa:
I have opened #881 which includes the changes I agree with as cherry-picked commits. Thank a lot for proposing them! I will close this PR, but your contribution will make it into the code base in #881.

PR #879, opened 2022-12-06
Reword some expressions

candleindark:
I have reworded some expressions. Again, the rewordings are subjective. Take them only if you see fit.


adswa:
Woah, thanks for all those PRs! :rocket: I will carve out some time to go through them today in the afternoon. As there are many changes within commits, I'll probably assemble the commits from all those many PRs into one PR, and within your commits I'll keep those changes I agree with. So once I've gotten the time to do that, you'll see your commits re-emerge in a new PR. :)

adswa:
I have opened #881 which includes the changes I agree with as cherry-picked commits. Thank a lot for proposing them! I will close this PR, but your contribution will make it into the code base in #881.

PR #895, opened 2022-12-16
Add a draft for annex encryption usecase

mslw:
This use demonstrates git-annex encryption in a workflow based on RIA stores and temporary clones.

This is based on a workflow we created for data incoming through a web form. A lot of credit for brainstorming the solution goes to @christian-monch, @bpoldrack, @aqw and others. I simplified a little, so that the workflow has two locations and can be easily divided into three parts.

The text covers the entire use, but can be seen as a rough draft in that it needs proofreading (both text style and DataLad correctness), formatting adjustment, and most likely extension/clarification of details. I won't be able to work on it in the next 2 weeks, so if anyone wants, please feel free to commit directly.

### DIFF
--- a/docs/usecases/encrypted_annex.rst
+++ b/docs/usecases/encrypted_annex.rst
@@ -1,0 +1,290 @@
+This use case demonstrates how to utilise `git-annex
+encryption <https://git-annex.branchable.com/encryption/>`__ to keep
+data encrypted when it is not being used. In this workflow, data need
+to be securely deposited on one machine, and downloaded to another
+machine for processing and storage. For this reason, in our workflow we
+will combine several independent pieces:
+
+#. Using git-annex encryption
+#. Using RIA stores
+#. Working in temporary (ephemeral) clones that combine information from
+   several remotes.
+
+The challenge
+=============
+
+Bob does data collection through a web form. Incoming form entries
+are saved as individual files by the web server. Bob periodically
+imports the data to his local machine, and uses them to produce an
+updated summary. For security reasons, he does not want the data to lie
+around unencrypted.
+
+The DataLad approach
+====================
+
+Bob sets up his remote machine (web server) so that incoming data is
+saved in a DataLad dataset. On that machine, he also creates a RIA
+sibling for the dataset (accessible from the outside through ssh), and
+enables encryption for annexed contents. He chooses a scheme in which
+the server only has the encryption key. Each new data file will be saved
+in the dataset, pushed to that RIA store, and dropped from the dataset,
+leaving only the encrypted copy on the remote server. On his local
+machine, he also sets up RIA stores, enables encryption, and scripts his
+analysis to check out the datasets locally, fetch updates from the
+remote machine, and push data back to local stores. In this way, data is
+not available anywhere in decrypted form, except when it is being
+processed.
+
+Step by step
+============
+
+Before we start: GnuPG
+----------------------
+
+DataLad relies on git-annex to manage large file content, and git-annex
+relies on `GnuPG <https://gnupg.org/>`__ to manage encryption. To
+proceed with the next steps, you will need at least one pair of GPG
+keys, one *private* (used for decryption) and one *public* (used for
+encryption). The relevant keys need to be known to the GPG program on
+the machine you are using.
+
+We won't go into detail about GPG, but the most useful commands are:
+``gpg --full-generate-key`` to generate a key pair, ``gpg --list-keys``
+and ``gpg --list-secret-keys`` to display known public and private keys,
+``gpg --export -a [keyID] > my_public_key.asc`` to export a public key
+to export a public key, and ``gpg --import my_public_key.asc`` to import
+the previously exported key on another machine.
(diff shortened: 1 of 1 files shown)


adswa:
This is a really cool write-up, thanks a lot @mslw for tackling this. I have pushed a few commits on top of yours already, and would be happy to see others giving it  a read as well.

docs/usecases/encrypted_annex.rst
```diff
@@ -0,0 +1,300 @@
...
+   $ datalad siblings -d inputs
+   .: here(+) [git]
+   .: origin(-) [/data/project/store/8e4/65aa4-af88-4abd-aaa0-d248339780be (git)]
+   .: localstore-storage(+) [ora]
+   .: entrystore-storage(+) [ora]
+
+Since we cloned the superdataset from local RIA store, also the subdataset has the `origin` (:term:`Git` :term:`remote`) pointing to that store.
+It also has the ``local-storage`` and ``entrystore-storage`` :term:`sibling`\s; these are the
+git-annex :term:`special remote`\s for the local and remote RIA stores, respectively.
+But to learn about new files that were added in the remote server since we last cloned from there, we need the Git remote.
```
mih, line 243: This will become unecessary https://github.com/datalad/datalad-next/issues/170

docs/usecases/encrypted_annex.rst
```diff
@@ -0,0 +1,300 @@
...
+If there were no updates reported, we could decide to finish our work
+right there. Since there are new files, we will integrate the changes
+(since we didn't change the input dataset locally, there is no practical
+difference in using ``ff-only`` versus ``merge``).
+
+.. code:: bash
+
+   $ datalad update --sibling entrystore --how merge
+
+.. find-out-more:: A note to users of python API
```
mih, line 271:
```suggestion
.. find-out-more:: A note to users of Python API
```

mih (commented):
This is great! Thx!

I left a bunch of comments, including style. In general, it would be nice to give capitalization of names another pass (pointed some out), and change the language to treat "data" as plural.

I link some issues to point out how this workflow would be simplified in the future. It might be worth including the links in the text, such that readers can find them until then.

mih:
One more: You could consider using `runrecord` directives to make sure that the code in this write-up continues to run.

adswa:
I was about to release the handbook, but realized that it would be best if this usecase would be a part of that release. Thus I will probably postpone a release into the next year.  In the new year, there are more eyes to review this, it is not pressing time wise to have a handbook release, and another advantage is that this would also give metalad the chance to get documented. This isn't meant as any time pressure, just an FYI :)

adswa (approved):
Thanks for the updates! I agree that the sample paths instead of full run records are sufficient to understand the usecase. During reading, I had the slight feeling that a sketch of the setup (of the involved places in the workflow) could be included for an easier overview of everything. In case you have one lying around from the SFB work you're doing, it would be cool to include it here. But if not, I think it also works fine without one - so it could be good to go from my side. I'll wait a bit with a merge to give @mih a chance for a re-review as well, though :)

docs/usecases/encrypted_annex.rst
```diff
@@ -0,0 +1,319 @@
...
+
+.. code:: bash
+
+   $ datalad create-sibling-ria \
+     --new-store-ok --name localstore \
+     --alias derived \
+     ria+file:///data/project/store
+   $ git annex enableremote \
+     localstore-storage \
+     keyid+=2EF857223E033AF1
```
mslw, line 180: I think it's missing encryption type declaration here & below - may be that hybrid is the default, but even if that's the case it's probably better to be explicit
```suggestion
   $ git annex enableremote \
     localstore-storage \
     encryption='hybrid' \
     keyid+=2EF857223E033AF1
```

mslw: Yes, I must have missed it. Fixed in 0668a40

mslw:
I agree with you @adswa that a better explanation of the setup was needed. I added two paragraphs at the beginning which should briefly and more clearly explain why we use "remote server" and "local machine" and why we set them up the way we do. I also fixed two `enableremote` commands where I forgot to add encryption mode (necessary).

adswa:
Thanks for the description, @mslw! I finally found the time to add an illustration to it as well. Would you mind checking if it is roughly how you would envision a sketch of the workflow?
I'm also giving @mih another ping here.

mslw:
The illustration looks good to me, thank you for preparing it!

For the record, this is what I had in mind when working on the thing:

PXL_20230118_172829676~2

But I think for the handbook your version works better. If I were to be picky, I would maybe swap places of the dataset and RIA within the cloud to highlight that the clone owned by the person on the right is made from RIA, and not the dataset checkout (i.e. form -> dataset -> RIA -> another dataset -> another RIA). And maybe make the small arrow between dataset and RIA on the right computer go both ways. But this is being picky, and the things can also stay as they are.

adswa:
As a heads-up: I'll merge this later today

PR #896, opened 2022-12-19
autorunrecord's main branch is now 'main'

mih:


adswa (approved):
thx! I'm pinging @brainhack-princeton on this as an FYI as they will want to make a similar change if they use ``autorunrecord``.

PR #898, opened 2022-12-19
Generic command output normalization

mih:
- Set a fixed identity and timestamp for Git (author and committer)
- Set a fixed seed for datalad UUID generation
- Remove all trailing spaces in captured output (avoids pertubations due to different Git configs across contributors)
- Replace virtual env location with 'VIRTUALENV' placeholder
- timestamp normalization in command output
- annex UUID standardization
- `ls -l` output normalization

Related: https://github.com/datalad/datalad/issues/7234

Outputs where captured with `sudo rm -rf /home/me/dl-101 /home/me/basics /home/me/pushes/ ; rm -f docs/basics/_examples/DL-101* ; make build`

### DIFF
--- a/docs/basics/101-101-create.rst
+++ b/docs/basics/101-101-create.rst
@@ -47,6 +47,7 @@ Let's start:
 .. runrecord:: _examples/DL-101-101-101
    :language: console
    :workdir: dl-101
+   :realcommand: ( mkdir DataLad-101 && cd DataLad-101 && git init && git config annex.uuid 46b169aa-bb91-42d6-be06-355d957fb4f7 ) &> /dev/null && datalad create --force -c text2git DataLad-101
    :cast: 01_dataset_basics
    :notes: Datasets are datalads core data type. We will explore the concepts of datasets by creating one with datalad create. optional configuration template and a description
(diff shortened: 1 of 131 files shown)


adswa:
This seems to work nicely . I'm thinking about adding the generated code snippets directly to this PR (the full build can complete over lunch) - what do you think?

adswa:
uhh I messed one up that's now missing. Need to push this in a short while (currently running a different build)

mih:
I pushed, what I believe could be the full change for basics 101

adswa:
Thank you!

adswa:
Oh man, I realized a problem this causes. This seems to normalize all dataset UUIDs? I spent hours trying to pinpoint why a create-sibling-ria call would fail, but then I saw that all datasets have identical UUIDs.

mih:
Ah! Good point. That is probably a sideeffect of fixing the random seed for datalad.

I think I will change this to support ENV overrides per `runrecord`, so we can fix the seed whenever needed and only then.

mih:
This is done now. A PR against the handbook is pending.

PR #900, opened 2022-12-20
Avoid a global random seed fix

mih:
This leads to all dataset IDs being identical.

This PR is not exhaustively annotating everything.

### DIFF
--- a/docs/basics/101-101-create.rst
+++ b/docs/basics/101-101-create.rst
@@ -47,6 +47,8 @@ Let's start:
 .. runrecord:: _examples/DL-101-101-101
    :language: console
    :workdir: dl-101
+   :env:
+     DATALAD_SEED=0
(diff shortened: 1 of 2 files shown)


adswa:
thanks. Do I understand correctly that every runrecord with a ``datalad create`` command needs this? or are there other commands as well that are influenced by the seed?

mih:
This only fixes the datalad seed. This means that all other randomness (e.g. in git-annex) is not affected (hence it has no impact on any of that functionality). Fixing the datalad seed for each runrecord that uses `create` makes sense for achieving the same dataset IDs. It would be important to use a different seed for each one (maybe the digits in the filename concatenated).

adswa:
Thanks a lot!

adswa (approved):

PR #901, opened 2022-12-20
Compartmentalize the Makefile , exit codes, and associated fixes

adswa:
This Changeset is quite messy - originally I wanted to merely split up the clean targets in the Makefile according to the book structure (2d6be3b415503eea4d74f3766e78f638a09c22f0). This brought up the need to change the working directories of two sections in the Advanced part (0a69b9a3133462138618ab006e544edafb71fb37, d85df9db21fa7aa142f13772f7a26a99d30bce7a). It also brought to light that identical dataset UUID generation causes problems in the RIA section (77fa11a2993958aaa1b7cb7d4b7083396b78136a). 
And because the fix in autorunrecord for the latter included a change that created a need to annotate exit codes, this PR also contains exit code annotations for commands that fail on purpose (c633936c692ab9588ffd5397e7c94184adca5028). It still isn't done (this PR is the code equivalent of a side quest), but I wanted to make the current state transparent.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -45,6 +45,8 @@ clean-beyond_basics:
 	@chmod +w -R /home/me/beyond_basics; rm -vrf /home/me/beyond_basics
 	# wipe out bare push repos
 	@chmod +w -R /home/me/pushes/data-version-control; rm -vrf /home/me/pushes/data-version-control
+	# wipe myriastore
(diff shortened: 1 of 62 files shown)


adswa:
I'll merge this now, and hope that the next PR will be less sweeping.

PR #913, opened 2022-12-21
Strip PATH listing from `wtf` when it has tmp dirs

mih:
Those have a random suffix and blow up the diff unneccessarily


mih:
Worked

PR #915, opened 2022-12-21
Add info on automatic CI builds to Contributing

### LINKED ISSUE #886: Add support for non-local builds
At the moment, a contribution to the handbook that adds a new code snippet requires the contributor to undergo a quite substantial technical setup (see e.g., https://github.com/datalad-handbook/book/issues/714).

This is because the book is only *assembled* automatically via CI, not  *build*. If we can change the workflow from "execute code snippets locally, commit captured results in ``_examples/*``, and push" to "execute code snippets in a github action and upload the results as artifacts", contributing would become slightly easier.

I suspect that the most promiment difficulty would lie in coordinating the execution of code snippets with the Sphinx build, because the Sphinx build relies on the existence of the code snippets to succeed.

adswa:
Adds info about the CI build option to the contributing page. Fixes #886

### DIFF
--- a/docs/contributing.rst
+++ b/docs/contributing.rst
@@ -23,6 +23,7 @@ to discuss changes or additions you plan to make in advance.
 Software setup
 ^^^^^^^^^^^^^^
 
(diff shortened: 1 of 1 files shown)


adswa:
in it goes (Appveyor was cancelled)

PR #918, opened 2022-12-22
Autobuild again

adswa:
For the unlikely chance that a new pr triggers an environment update


adswa:
in it goes! thanks @mih!

Issue #824, opened 2022-04-08
Replace hardcoded reference(s) to figure numbers


mslw:
Following PR #820 I realised that I should have suggested `:numref:`. Worth checking if there are more hardcoded references to figure numbers, but a cursory glance suggests that it might have been the only one.

There aren't many references to figures in text overall; an example of `:numref:` usage can be found in `docs/usecases/datastorage_for_institutions.rst`

PR #820, opened 2022-04-07
fix wrong figure name

jhpb7:
The figure "names" was wrong. I changed it to the correct one.

PR #919, opened 2022-12-22
BF: replace hardcoded figure number with numref

jsheunis:
This PR closes https://github.com/datalad-handbook/book/issues/824

Looks like the hardcoded figure number (6.2, from https://github.com/datalad-handbook/book/pull/820/commits/d1033b5c5e81839f4a85f9695f0ff1f5ebb42d40) was also incorrect (if I read the context correctly) - it should be Fig 6.1.

I searched for other instances of hardcoded figure numbers in the book but couldn't find any.


adswa (approved):
Thanks much! I'll cancel the Appveyor build as the handbook build already succeeded, and there isn't any changed code :)

PR #923, opened 2023-01-06
BF: fix invocation of siblings to configure publish-depends

yarikoptic:


adswa:
Thanks for the PR. This can go in from my pov, the readthedocs failure looks as if you pr'ed from an older state where autorunrecords default branch was still called master? (On my phone, can't check easily). I think it should be fine upon a rebuild once it's merged

adswa (approved):

adswa:
Cancelled the appveyor run as there was no run record change

PR #925, opened 2023-01-12
Make a note that content stays in the annex after unannexing

### LINKED ISSUE #924: Mention the necessity to drop when unannexing files
I realized when reading git-annex unannex's manpage and a usecase about an accidentially annexed file that was already pushed to a special remote that accidentally annexed large files need to be dropped still.

```
adina@muninn in ~
❱ git annex help unannex  
git-annex-unannex(1)                         General Commands Manual                        git-annex-unannex(1)
 
NAME
       git-annex-unannex - undo accidental add command
 
SYNOPSIS
       git annex unannex [path ...]
 
DESCRIPTION
       Use this to undo an accidental git annex add command. It puts the file back how it was before the add.
 
       Note  that  for  safety, the content of the file remains in the annex, until you use git annex unused and
       git annex dropunused.
 
       This is not the command you should use if you intentionally added a file some time ago,  and  don't  want
       its  contents  any  more. In that case you should use git annex drop instead, and you can also git rm the
       file.
 
```

This needs to be mentioned in the file system operations chapter.

adswa:
Fixes #924

### DIFF
--- a/docs/basics/101-136-filesystem.rst
+++ b/docs/basics/101-136-filesystem.rst
@@ -865,7 +865,7 @@ is only two files, and they are quite small, you decide to store them in Git --
 this way, the files would be available without configuring an external data
 store.
 
-To get contents out of the dataset's annex you need to *unannex* them. This is
+To get a file out of the git-annex hands you need to *unannex* it. This is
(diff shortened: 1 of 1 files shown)


adswa:
Cancelled the appveyor run; no code snippet changes

PR #926, opened 2023-01-16
Make analytics stats work with any RTD project

mih:
One can/must now pass a project name as a positional argument


adswa:
Cancelled the Appveyor run

adswa (approved):

PR #935, opened 2023-02-26
Update the DVC section

mslw:
Minor fixes for things I noticed when preparing for a workshop. See commit messages for details.

Lacks run records, which prevents merge, but depends on appveyor succeeding (see below), so I leave this in draft mode. Otherwise quite complete.

Remaining todos and issues (help welcome):

- [x] Update run records. Some of the commits touch the `runrecord` directives and they will require updating the built [run records](https://handbook.datalad.org/en/latest/contributing.html#automatic-builds), preferably with artifacts from AppVeyor once it succeeds (see comment below).
- [x] Remove git errors when pushing tags. Currently, two outputs in the DVC workflow (ctrl-f in [this section](https://handbook.datalad.org/en/latest/beyond_basics/101-168-dvc.html#id13)) contain "error" and "fatal" messages coming from git. My first impression is that they are caused by being unable to push tags that already exist (but probably are attached to different commit ids). Might be a holdover from some previous run, or a cleaning error? There are some `.. only:: adminmode` directives that create a local push target to serve as origin, after all. (nothing critical by any means)

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -27,7 +27,8 @@ environment:
   DTS: datalad_next
   APPVEYOR_BUILD_WORKER_IMAGE: Ubuntu2004
   INSTALL_SYSPKGS: python3-virtualenv graphicsmagick-imagemagick-compat moreutils jq uidmap
-  INSTALL_GITANNEX: git-annex -m snapshot
+  INSTALL_GITANNEX: git-annex -m deb-url --url http://snapshot.debian.org/archive/debian/20210906T204127Z/pool/main/g/git-annex/git-annex_8.20210903-1_amd64.deb
+
 
 
 skip_commits:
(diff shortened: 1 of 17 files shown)


mslw:
Appveyor error seems to be https://github.com/datalad/datalad/issues/7286 - looks like I'll be building locally 🙃

adswa (commented):
Oh wow, you found quite a bunch of problems! Thanks much for digging through this. I'll see if I can get Appveyor to work again :+1:

adswa:
As for:

>     * [ ]  Remove git errors when pushing tags. Currently, two outputs in the DVC workflow (ctrl-f in [this section](https://handbook.datalad.org/en/latest/beyond_basics/101-168-dvc.html#id13)) contain "error" and "fatal" messages coming from git. My first impression is that they are caused by being unable to push tags that already exist (but probably are attached to different commit ids). Might be a holdover from some previous run, or a cleaning error? There are some `.. only:: adminmode` directives that create a local push target to serve as origin, after all. (nothing critical by any means)

I believe you are referring to error such as this one, right? 

```
error: src refspec sgd-pipeline matches more than one
error: failed to push some refs to '/home/me/pushes/data-version-control'
```

First of all, thanks for catching them! I believe the problem is that there is both a tag and a branch with the same name (``sgd-pipeline`` and ``random-forest``, and (I think) this is due to the fact that they exist on GitHub... I have removed them upstream, and triggered another appveyor build. Let's see :crossed_fingers:

adswa:
Alright, I have addressed your TODOs @mslw. I'm marking this as ready.

adswa (approved):

PR #943, opened 2023-04-18
Mention git annex uninit in FAQ and file system operations

### LINKED ISSUE #939: Add git annex uninit to the handbook
The origin is this use case by @Remi-Gau from the office hour chat: 

> hey y'all
I created a datalad dataset that is in the end just used as a super-dataset to link to a bunch of sub-datasets: so basically I could just use a simple git repo for that.
Is there any way to "remove" the annex entirely: sure I can remove the git-annex branch, but annex things will still be there in the .gitfolder.

follow up:
> not that it does remove things locally but doing a datalad push after that brings the branch back to life.
had to remove the the git annex branch on the sibling to prevent that from happening

adswa:
Fixes #939


adswa:
cancelled Appveyor builds since no code snippets need to be generated.

PR #945, opened 2023-04-18
Add autoenable=true to Gitlfs special remote

### LINKED ISSUE #844: autoenable=true the github-lfs remote?
somewhat a continuation to #841

shouldn't we add `autoenable=true` to `initremote` in http://handbook.datalad.org/en/latest/basics/101-139-gitlfs.html ? or at least add a hint dropdown about it?  we do have it in http://handbook.datalad.org/en/latest/basics/101-139-s3.html?highlight=autoenable#initialize-the-s3-special-remote

adswa:
Fixes #844

### DIFF
--- a/docs/basics/101-139-gitlfs.rst
+++ b/docs/basics/101-139-gitlfs.rst
@@ -11,19 +11,19 @@ quite easily.
 
 In order to store annexed dataset contents on GitHub, we need first to create a repository on GitHub::
 
-    $ datalad create-sibling-github test-github-lfs
-    .: github(-) [https://github.com/yarikoptic/test-github-lfs.git (git)]
-    'https://github.com/yarikoptic/test-github-lfs.git' configured as sibling 'github' for <Dataset path=/tmp/test-github-lfs>
+    $ datalad create-sibling-github test-github-lfs --access-protocol ssh
+    .: github(-) [git@github.com:yarikoptic/test-github-lfs.git (git)]
+    'git@github.com:yarikoptic/test-github-lfs.git' configured as sibling 'github' for <Dataset path=/tmp/test-github-lfs>
 
 and then initialize a :term:`special remote` of type ``git-lfs``, pointing to the same GitHub repository::
 
-    $ git annex initremote github-lfs type=git-lfs url=https://github.com/yarikoptic/test-github-lfs autoenable=true encryption=none embedcreds=no
+    $ git annex initremote github-lfs type=git-lfs url=git@github.com:yarikoptic/test-github-lfs autoenable=true encryption=none embedcreds=no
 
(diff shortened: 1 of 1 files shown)


yarikoptic:
I guess if that is an intention to have it autoenabled -- it could be added, but then may be text should be adjusted accordingly to describe that?  May be it should be just added as yet another "option" (like `compression`), e.g. "If you would like this LFS special remote to get automatically enabled whenever dataset is cloned, add `autoenable=true` to the `initremote` invocation above", or smth like that?

adswa:
Since it was your issue I'm trying to fix, which of those suggestions would you like to have? ;-)

yarikoptic:
rereading the text I noted in https://github.com/datalad-handbook/book/pull/945/files#diff-c2337bf386e1a4929286b4f237683208f737edc975cf5af59dd64fd5f4599d5eL47


For consumers of your dataset, because the special remote URL coincides with the regular remote URL on GitHub, ``siblings enable`` (as shown in the section :ref:`dropbox`) will not even be necessary when datalad is installed from GitHub.

so, if that is the true, then we do not even need to add `autoenable=true` and the fix would be to follow up with that knowledge on that issue/question and have it  closed.  Would you be so kind to try to follow this recipe and see if it is indeed true?

adswa:
I tried, this is what I found:

- ``git annex copy`` only succeeds when the special remote is registered with an SSH URL
- If I clone the dataset again from github the special remote is not autoenabled, even though the URLs of special remote and git remote are identical:
```
(handbook) adina@muninn in /tmp
❱ datalad clone git@github.com:adswa/test-github-lfs.git
[INFO   ] Unable to parse git config from origin                                
[INFO   ] Remote origin does not have git-annex installed; setting annex-ignore 
|   This could be a problem with the git-annex installation on the remote. Please make sure that git-annex-shell is available in PATH when you ssh into the remote. Once you have fixed the git-annex installation, run: git annex enableremote origin 
[INFO   ] access to 1 dataset sibling github-lfs not auto-enabled, enable with:
| 		datalad siblings -d "/tmp/test-github-lfs" enable -s github-lfs 
install(ok): /tmp/test-github-lfs (dataset)
```
```
(handbook) adina@muninn in /tmp/git-lfs-test on git:master
❱ git remote -v
github	git@github.com:adswa/test-github-lfs.git (fetch)
github	git@github.com:adswa/test-github-lfs.git (push)
github-lfs	git@github.com:adswa/test-github-lfs.git (fetch)
github-lfs	git@github.com:adswa/test-github-lfs.git (push)
```

I'm adjusting the PR to use an SSH url, add ``autoenable=true``, and remote the paragraph about autmatic autoenabling.

PR #946, opened 2023-04-18
Cross-ref KBI content on interactive computing with datalad run

### LINKED ISSUE #805: `datalad run` chapter: more information/examples
### Is your feature request related to a problem?

Just read the datalad run chapter and get direct questions:

- should the git status be empty before running `datalad run` or will datalad run only save the files created during `datalad run` even if other changes were made before running the code ?
- Can datalad run R scripts, too. If yes, we probably need to run the command in the "project directory" when using RStudio, right? (which may be the code directory)

### Describe the solution you'd like

extend the chapter to  answer these questions

### Describe alternatives you've considered

get a chapter on using datalad with R/Rstudio, (if existent but I did not get to it yet, link it here)

### Additional context

_No response_

adswa:
Fixes #805


adswa:
Cancelled the appveyor build as no new code snippets need to be generated

PR #951, opened 2023-05-02
Small grammar error fixed

cs-hall:


adswa (approved):
Thanks for catching this!

PR #958, opened 2023-05-19
two small errors fixed

cs-hall:
Two minor errors I found while reading


cs-hall:
Is it better to create a different pull request for each edit or collect them in batches and submit one request for many small changes?

PR #961, opened 2023-05-31
Added section for issues arising when using basic shell commands

TheChymera:
I'm regularly experiencing issues when pointing basic shell commands at datalad datasets, these incompatibilities should be documented somewhere in a clear and transparent fashion. Draft for now, as I re-encounter more of these issues.

@yarikoptic

### DIFF
--- a/docs/basics/shell.rst
+++ b/docs/basics/shell.rst
@@ -1,0 +1,52 @@
+.. _shell:
+
+Using Common Shell Commands on DataLad Repositories
+---------------------------------------------------
+
+There are a number of gotchas when using common shell commands to manipulate DataLad Repositories.
+This stems mostly due to the reliance of git-annex on symlinks, as well as due to permissions management under the ``.git`` directory.
+Thus, the usage of common shell tools, including `GNU coreutils <https://www.gnu.org/software/coreutils/>`_, can often yield unintuitive results.
+This is a list referencing such cases, providing instructions on how to mitigate them, as well as indicating alternative processes using datalad-specific commands, where such are feasible substitutes.
+
+Removal via ``rm``
+~~~~~~~~~~~~~~~~~~
+
+Problem
+^^^^^^^
+
+Removing datalad repositories with ``rm`` will fail due to missing write permissions under ``.git/annex/objects/..``:
+
+.. code-block::
+   [deco]/tmp ❱ datalad clone https://github.com/datalad/testrepo--basic--r1
+   [INFO   ] Remote origin not usable by git-annex; setting annex-ignore
+   [INFO   ] https://github.com/datalad/testrepo--basic--r1/config download failed: Not Found
+   install(ok): /tmp/testrepo--basic--r1 (dataset)
+   [deco]/tmp ❱ cd testrepo--basic--r1/
+   [deco]/tmp/testrepo--basic--r1 ❱ datalad get .
+   get(ok): test-annex.dat (file) [from web...]
+   [deco]/tmp/testrepo--basic--r1 ❱ cd ..
+   [deco]/tmp ❱ rm -rf testrepo--basic--r1/
+   rm: cannot remove 'testrepo--basic--r1/.git/annex/objects/zk/71/SHA256E-s4--181210f8f9c779c26da1d9b2075bde0127302ee0e3fca38c9a83f5b1dd8e5d3b.dat/SHA256E-s4--181210f8f9c779c26da1d9b2075bde0127302ee0e3fca38c9a83f5b1dd8e5d3b.dat': Permission denied
+
+Solution
+^^^^^^^^
+
+This can be mitigated by manually changing permissions:
+
+.. code-block::
+   [deco]/tmp ❱ chmod -R 775 testrepo--basic--r1/
+   [deco]/tmp ❱ rm -rf testrepo--basic--r1/
+
+Alternative
+^^^^^^^^^^^
+
+DataLad provides its own command to substitute for ``rm`` functionality:
+
+.. code-block::
+   [deco]/tmp ❱ datalad remove -d testrepo--basic--r1/
+   drop(ok): . (key)
+   uninstall(ok): . (dataset)
+   action summary:
+     drop (ok: 1)
+     uninstall (ok: 1)
+


adswa:
Hey, thanks for the PR! 

There is a handbook section that touches upon the topic you bring up: https://handbook.datalad.org/en/latest/basics/101-136-filesystem.html. It also has a paragraph on ``rm -rf`` of entire datasets: https://handbook.datalad.org/en/latest/basics/101-136-filesystem.html. I think it would make sense to merge the changes you propose here into the section that already exist where they fit, e.g., the existing section uses ``chmod u+w -R`` instead of octal modes -- worthwhile to include your command as an alternative, and its actually also nice that you have a full example including cloning a dataset. What do you think?

docs/basics/shell.rst
```diff
@@ -0,0 +1,52 @@
...
+   [INFO   ] Remote origin not usable by git-annex; setting annex-ignore
+   [INFO   ] https://github.com/datalad/testrepo--basic--r1/config download failed: Not Found
+   install(ok): /tmp/testrepo--basic--r1 (dataset)
+   [deco]/tmp ❱ cd testrepo--basic--r1/
+   [deco]/tmp/testrepo--basic--r1 ❱ datalad get .
+   get(ok): test-annex.dat (file) [from web...]
+   [deco]/tmp/testrepo--basic--r1 ❱ cd ..
+   [deco]/tmp ❱ rm -rf testrepo--basic--r1/
+   rm: cannot remove 'testrepo--basic--r1/.git/annex/objects/zk/71/SHA256E-s4--181210f8f9c779c26da1d9b2075bde0127302ee0e3fca38c9a83f5b1dd8e5d3b.dat/SHA256E-s4--181210f8f9c779c26da1d9b2075bde0127302ee0e3fca38c9a83f5b1dd8e5d3b.dat': Permission denied
+
```
adswa, line 30: The ``.. code-block::`` directives currently cause a build error because they need a space between the directive and the code:
```suggestion
.. code-block::

   [deco]/tmp ❱ datalad clone https://github.com/datalad/testrepo--basic--r1
   [INFO   ] Remote origin not usable by git-annex; setting annex-ignore
   [INFO   ] https://github.com/datalad/testrepo--basic--r1/config download failed: Not Found
   install(ok): /tmp/testrepo--basic--r1 (dataset)
   [deco]/tmp ❱ cd testrepo--basic--r1/
   [deco]/tmp/testrepo--basic--r1 ❱ datalad get .
   get(ok): test-annex.dat (file) [from web...]
   [deco]/tmp/testrepo--basic--r1 ❱ cd ..
   [deco]/tmp ❱ rm -rf testrepo--basic--r1/
   rm: cannot remove 'testrepo--basic--r1/.git/annex/objects/zk/71/SHA256E-s4--181210f8f9c779c26da1d9b2075bde0127302ee0e3fca38c9a83f5b1dd8e5d3b.dat/SHA256E-s4--181210f8f9c779c26da1d9b2075bde0127302ee0e3fca38c9a83f5b1dd8e5d3b.dat': Permission denied
```

However, what I would suggest we do is transforming them into a ``.. autorunrecord::`` directive. These get executed automatically by appveyor and will always show how the current software stack currently behaves, i.e., they are far better maintainable. 

```suggestion
.. runrecord:: _examples/DL-101-136-185
   :language: console
   :workdir: dl-101
   :exitcode: 1
   
   $ datalad clone https://github.com/datalad/testrepo--basic--r1
   $ cd testrepo--basic--r1/
   $ datalad get .
   $ cd ..
   $ rm -rf testrepo--basic--r1/
```

Any line starting with a ``$`` will be executed in sequence in the ``dl-101`` workdir (on Appveyor), and the output of all commands will be written to ``_examples/DL-101-136-185``. The code block expects a non-zero exit code (I hope 1 is the correct exit code, I didn't try). The [Contributing](https://handbook.datalad.org/en/latest/contributing.html) page has a dedicated section that explains the technical backbone behind it.

TheChymera: @adswa I saw that in the other examples, the issue with that is the prompt doesn't show the current working directory, which makes things a bit unclear. Is there any way to do that?

Also:

> The code block expects a non-zero exit code 

given the nature of this section, a lot of the commands are supposed to fail, so “proper” examples of what fails might show up as errors.

adswa: > @adswa I saw that in the other examples, the issue with that is the prompt doesn't show the current working directory, which makes things a bit unclear. Is there any way to do that?

In theory, we could configure it like this, but a cheap way to do it is to add a comment line (``#`` prefix) where you simply state "Outside of the dataset".  

> given the nature of this section, a lot of the commands are supposed to fail, so “proper” examples of what fails might show up as errors.

Sorry, I should have been more clear on this. Whenever a runrecord is meant to fail for whatever reason (we often show errors on purpose), it needs this annotation. Otherwise, the CI system treats it as a failure.

adswa:
After a long while, hi again! I've been a bit busy with many other projects, but rediscovered this draft PR. I have restarted the appveyor run. I saw that the PR is in Draft mode, do you have plans to eventually take it out of draft mode, or is it meant to be taken over by someone else?

docs/basics/shell.rst
```diff
@@ -0,0 +1,185 @@
...
+~~~~~~~~~~~~~~~~~~
+
+Problem
+^^^^^^^
+
+Removing datalad repositories with ``rm`` will fail due to missing write permissions under ``.git/annex/objects/..``:
+
+.. runrecord::
+   :language: console
+   :exitcode:1
```
adswa, line 21:
```suggestion
   :exitcode: 1
```

docs/basics/shell.rst
```diff
@@ -0,0 +1,185 @@
...
+   rm: cannot remove 'testrepo--basic--r1/.git/annex/objects/zk/71/SHA256E-s4--181210f8f9c779c26da1d9b2075bde0127302ee0e3fca38c9a83f5b1dd8e5d3b.dat/SHA256E-s4--181210f8f9c779c26da1d9b2075bde0127302ee0e3fca38c9a83f5b1dd8e5d3b.dat': Permission denied
+
+Solution
+^^^^^^^^
+
+This can be mitigated by manually changing permissions:
+
+.. runrecord::
+   :language: console
+   :exitcode:0
```
adswa, line 42:
```suggestion
   :exitcode: 0
```

docs/basics/shell.rst
```diff
@@ -0,0 +1,185 @@
...
+
+Solution
+^^^^^^^^
+
+This can again be compensated for by explicitly instructing the command to follow symlinks:
+
+.. code-block::
+
+   [deco]/mnt/data/testrepo--basic--r1 ❱ du -sL myfile
+   1024	myfile
```
mslw, line 185: 100 % agree that `du -sL` is the answer here, but I wonder if it's a right place to drop a hint that `git annex info` also provides size information for annexed files or directories -- although the *disk usage* is less direct, e.g for file it reports file size and whether it is present or not.

mslw (commented):
Hi @TheChymera 

I really like this addition, and I learned a few things from it. I think it would be great to get it merged.

I realize that we let this PR approaching a year in a draft state :grimacing: , Now that I looked into it closely, I think it would require some tweaks, mostly technical. Please let me know whether you'd be interested in finishing it, or whether you'd prefer me to take over - in either case I would be ready to assist.

I left notes throughout the document (in the latter case, they can be "note to self"). I also have some general comments below.

- I wonder about the placement for this segment. I think the Appendix (as currently proposed) is ok if a bit out of sight. Looking at the table of content, maybe it could slot into basics chapter 9, or potentially elsewhere?

- I am thinking about run records. There are few DataLad commands here, and having their output tested does not seem critical; but if they can be easily wrapped into run records for consistency then why not. Some complexity may be related to the fact that we need to set up then tear down a dataset a few times to demonstrate everything. Most of the examples focus on non-DataLad commands, and these probably do not require run records (we could install ripgrep and ack into the test env, but there seems to be little point in verifying outputs of those).

- Related, I do not think that keeping the working directory in the prompt is necessary to make the examples understandable. There isn't that much jumping with `cd`, and when needed, bash comments can be used. 
  - there seems to be a virtual env name in the prompt (`[deco]`) which seemed distracting to me, I think it could be removed
  - if code blocks (and not run records) are used, and a prompt is reduced to `$`, we can declare them as `code-block:: console` (or shell-session), in which case BashSessionLexer would provide nice coloring

PR #966, opened 2023-06-15
Update 101-164-dataladdening.rst

jcolomb:
the original workflow do not work.

### DIFF
--- a/docs/beyond_basics/101-164-dataladdening.rst
+++ b/docs/beyond_basics/101-164-dataladdening.rst
@@ -52,7 +52,7 @@ Consider :ref:`applying procedures <procedures>` with ``-c <procedure-name>`` to
    If you want to transform a Git repository to a DataLad dataset, a :command:`datalad create -f` is the way to go, too, and completely safe.
    Your Git history will stay intact and will not be tampered with.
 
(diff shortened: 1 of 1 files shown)


docs/beyond_basics/101-164-dataladdening.rst
```diff
@@ -52,7 +52,7 @@ Consider :ref:`applying procedures <procedures>` with ``-c <procedure-name>`` to
    If you want to transform a Git repository to a DataLad dataset, a :command:`datalad create -f` is the way to go, too, and completely safe.
    Your Git history will stay intact and will not be tampered with.
 
-If you want to transform a series of nested directories into nested datasets, continue with :command:`datalad create -f` commands in all further subdirectories.
+If you want to transform a series of nested directories into nested datasets, you will need to use :command:`datalad create -f -d` commands giving the path of the nested directory. (If you are using the create function inside the folders, the superdataset will not treat them a subdatasets.)
```
jcolomb, line 55:
```suggestion
If you want to transform a series of nested directories into nested datasets, you will need to use :command:`datalad create -f -d .` commands giving the path of the nested directory. (the `-d .` option tells datalad to add the dataset as a subdataset to the supradataset (`-d`) found in the current directory (`.`).
```

adswa:
huh, thats interesting, thanks for noticing and the PR! I need to check if this is a bug in datalad.

PR #967, opened 2023-06-19
Update installation.rst for GIt for windows install

jcolomb:
delete option that do not exist anymore
add option about openssh, as of  https://github.com/datalad/datalad/issues/5829#issuecomment-1068773537


adswa (approved):
thx!

PR #969, opened 2023-07-21
Removing cmdline prompt character installation.rst

Arshitha:
Removing cmdline prompt character (>) from the git-annex section so it can be copy-pasted into cmdline and without making any edits. As it is currently, user would have to manually delete the > character before the command runs. Super minor change but thought it might be useful for others

### DIFF
--- a/docs/intro/installation.rst
+++ b/docs/intro/installation.rst
@@ -125,13 +125,13 @@ Git-annex:
 
     .. code-block:: bat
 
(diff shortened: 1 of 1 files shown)


adswa (approved):
Thank you very much!  :heart:

adswa:
hey @all-contributors please add @Arshitha for bug, maintenance

PR #993, opened 2023-08-07
Add an external link to the redirection list

mih:


adswa:
I see that this PR is in draft mode, can I help to get it out of it?

PR #996, opened 2023-08-07
Robustify linkchecker workflow

mih:
Was failing due to outdated package lists.

Also try to speed up installation with `eatmydata`


mih:
I will merge this and take care of any fallout. Needs to be tested from `main`

PR #999, opened 2023-08-07
Update DVC workflow for DVC v3

### LINKED ISSUE #965: Update DVC use case for DVC 3+
It looked like the DVC section finally broke, Appveyor isn't happy. I need to go through the DVC walkthrough and update it to the latest commands, or pin the DVC version I think

mslw:
DVC 3.0 was released with a set of breaking changes ([release note](https://github.com/iterative/dvc/releases/tag/3.0.0)). However, the only one which affects us is the removal of `dvc run` in favour of `dvc stage add --run`. This commit updates the DVC workflow to use the 3.0 command, and removes related build failures to resolve #965.

This is effectively a reversal of 47b3ed9 which pinned dvc < 3.0 for our build. While there were multiple quick-release 3.x versions (going from 3.0 to 3.12 in the course of 3 weeks), they seem to adhere to semver, as the changelog reports no breaking changes. At a quick glance, none of the new features or bug fixes are particularly relevant for our (rather basic) workflow, so it seems OK to me to go with an updated major version dependency.


adswa (approved):
Thanks a lot, and sorry for missing this for so long! I have the typos that make codespell fail fixed in main already.

PR #1016, opened 2023-08-10
add extra instruction for windows users (installation

jcolomb:
I think it would be good to add some tests for other steps.
Alternatively or in addition, one could have a debug section with the tests at the end ?

People tend to go to fast and make mistakes, debugging the installation makes little sense (starting over is simpler).

### DIFF
--- a/docs/intro/installation.rst
+++ b/docs/intro/installation.rst
@@ -75,6 +75,10 @@ There are countless ways to install software on Windows. Here we describe *one*
 possible approach that should work on any Windows computer, like one that you
 may have just bought.
 
+Unless you know what you are doing, try to follow the steps precisely. 
+If something is not working, you probably did something wrong in the previous step.
+It is often faster and easier to start over than trying to find out what went wrong.
+
 
 Python:
     Windows itself does not ship with Python, it must be installed separately.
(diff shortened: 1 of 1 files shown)


jcolomb:
@adswa I would work this PR if you think it makes sense, what do you think about adding tests at each step ?

docs/intro/installation.rst
```diff
@@ -75,14 +75,18 @@ There are countless ways to install software on Windows. Here we describe *one*
...
+
 
 Python:
     Windows itself does not ship with Python, it must be installed separately.
     If you already did that, please check the :find-out-more:`on Python
     versions <fom-py2v3>`, if it matches the requirements. Otherwise, head over
     to the `download section of the Python website
     <https://www.python.org/downloads>`_, and download an installer. Unless you
-    have specific requirements, go with the 64bit installer of the latest
+    have specific requirements, go with the 4bit installer of the latest
```
mih, line 89: This change should be removed.

docs/intro/installation.rst
```diff
@@ -96,6 +100,8 @@ Python:
     as this is required for subsequent installation steps and interactive use later on.
     Other than that, using the default installation settings is just fine.
 
+    If you did everything right, typing ``python --version`` in the command prompt should give you an answer.
```
JoKeyser, line 103: Here you could mention the kind of correct answer (instead of an "answer" like "command not found"), e.g. "should show you your installed Python version, like 'Python 3.x.y'."

mih:
We have a much more simplified way to install datalad/git-annex on windows. It no longer requires any manual installation of Python. I believe this PR can be closed for good.

Thanks for the initiative!

Issue #1009, opened 2023-08-09
PDF version issues


mih:
This is a summary issue, listing open issues in the PDF version. The described state is after https://github.com/datalad-handbook/book/pull/1003 (which solved most "overfull" and line-wrapping issues. Moreover, this is focused on a 6x9" paper PDF.

- [x] long footnote URL: https://git-annex.branchable.com/projects/datalad/bugs-done/Windows__58___substantial_per-file_cost_for___96__add__96__ EDIT: with the improvements in index generation, this footnote fits a single line
- [ ] long example code URL: https://sourceforge.net/projects/linuxcommand/files/TLCL/19.01/TLCL-19.01.pdf/download
- [x] two different URLs for byte-of-python (code vs reference)
- [ ] long-now MP3 listing has loooong filenames
- [ ] long-now root folder loooong ` recordings/longnow/Long_Now__Seminars_About_Long_term_Thinking`
- [x] `less recordings/podcasts.tsv` needlessly long (1/3 page), could be 4-5 lines
- [x] in many places/listing `../.git/annex/objects` still takes up a lot of redundant space. Maybe it could be cut and replaced wholesale together with the keydir. We could leave some kind of symbol instead
- [x] #1000 (for asciicast URLs)
- [x] long listing of `$ git config --list --show-origin` (half a page), could be just a few lines
- [x] we should consider turning off the "action summary". This is useful for commands with MANY results, we we rarely (ever?) have that in the handbook. The current setup essentially near-doubles the number of lines in a short listing.
- [x] spacing issue for `(annex_largefiles='((mimeencoding=binary)and(largerthan=0))'`
- [x] excessive box `M10.6 Write your own procedures` (4 pages). The containing sections hardly has anything else. The box is not references elsewhere. It should simply not be a box, but become the main content of that section.
- [x] Reformat `code/script.py` in YODA analysis to shorten line lengths
- [x] Try shrink `M11.6 What is GitHub?` to fit on a single page (possibly by trimming the screenshots
- [ ] Excessive boxes (multipage length, unreferenced, nothing else as content in section, could be the actual section content)
  - [x] `M11.2 DataLad’s Python API`
  - [ ] `M11.3 Creating an independent input dataset`
  - [x] `M11.7 On the looks and feels of this published dataset`
  - [x] `M12.1 More on datalad status`
  - [ ] `M14.4 If a renamed/moved dataset is a sibling. . .`
- [x] Large screenshots (could simply bit cut or scaled)
[...]

PR #1021, opened 2023-08-11
Turn the "eternal" handbook date into symbols

mih:
We fix the committer date in the handbook for reasons of reproducibility (no random shasum changes, simply because we execute code at a different time). Hence the dates/times shown do not really make sense (they appear constant, even across multiple steps of a procedure. This appears odd.

This changeset replace the timestamps with symbolic placeholders.

Their meaning still needs to be described in the intro.

Ping #1009

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -34,6 +34,9 @@ autorunrecord_basedir = '/home/me'
 autorunrecord_line_replace = [
     # trailing space removal
     (r'[ ]+$', ''),
+    # turn our "eternal" date time into symbols
+    (r'Tue Jun 18 16:13:00 2019 +0000', '🗓🕓'),
+    (r'2019-06-18 16:13', '🗓🕓'),
     # strip the keydir for MD5(E) or SHA1(E) annex keys
     # the keydir is identical to the annex key name, but consumes
     # a lot of space. we replace it with a UTF scissors icon
--- a/docs/latex/fontpkg.sty
+++ b/docs/latex/fontpkg.sty
@@ -4,6 +4,8 @@
 \usepackage{microtype}
 
 % provide definition for Unicode character ✂ (U+2702) we use for truncation
-\usepackage{marvosym}
+\usepackage{fontawesome}
 \usepackage{newunicodechar}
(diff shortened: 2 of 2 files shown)


mih:
The regex does not have the desired effect. Unclear why.

docs/conf.py
```diff
@@ -34,6 +34,9 @@
 autorunrecord_line_replace = [
     # trailing space removal
     (r'[ ]+$', ''),
+    # turn our "eternal" date time into symbols
+    (r'Tue Jun 18 16:13:00 2019 +0000', '🗓🕓'),
+    (r'2019-06-18 16:13', '🗓🕓'),
```
mslw, line 39: I see two reasons why the regexp would not work:
- The plus symbol needs to be escaped (otherwise it's "space, one or more, followed by 0000")
- The time zone can not be taken for granted, the config sets git author & committer date to `2019-06-18T16:13:00`, and for me this produces run records with git log saying `Tue Jun 18 16:13:00 2019 +0200` (I'm in CEST).

I'm also adding a space for IMHO better aesthetics.


```suggestion
    (r'Tue Jun 18 16:13:00 2019 \+\d{4}', '🗓 🕓'),
    (r'2019-06-18 16:13', '🗓 🕓'),
```

mslw, line 39: As a side note, I would use 📅 (Calendar, [U+1F4C5](https://unicodeplus.com/U+1F4C5)) instead of 🗓 (Spiral Calendar Pad, [U+1F5D3](https://unicodeplus.com/U+1F5D3)), but that's a personal preference and likely depends on my set of browser fonts (?) -- not leaving a GitHub suggestion as this would require a one-line change to fontpkg.sty, too.

mslw (commented):
I think I figured out the regexp @mih (see inline comments), so if you are still interested in having the replacement, then we can go ahead. A TODO from the commit message ("their meaning still needs to be described in the intro") still applies.

FTR, this is what `git log` run record from 101-101-create looks like in my browser:

Screenshot from 2024-05-21 11-31-25

PR #1022, opened 2023-08-11
codespell: move configuration into pyproject.toml and fix "re-use" typo

yarikoptic:
so we could just run `codespell` and not rely on config being only github action specific


adswa (approved):
thx!

PR #1032, opened 2023-10-09
Update (windows) installation instructions

### LINKED ISSUE #1031: Update/enhance windows installation instructions
Right now, this is a collection of notes. I am not yet sure in what form this could translate into an update of the handbook.

#### Non-admin Python installation

Yes, it is possible, and it is the installer default.

#### Conflicting Python versions

There could already be Python version installed. For example, the `inkscape` installer includes one and it also adds it in the PATH via its own `bin/` directory.

To check for possible installations, open a terminal (`cmd.exe`) and run `where python`.

If there are multiple versions, run the following to see which one is the default: `python -c "import sys; print(sys.executable)"`

If this is not the version you expected or installed, you may need to adjusted the PATH environment variable setting to put the installation directory of the desired version before any other.

#### Install git-annex via datalad-install

This should use the `python -m pip install datalad-installer` syntax, to make sure we use the python environment that we selected.

#### Install datalad via pip

This should use the `python -m pip install datalad` syntax, to make sure we use the python environment that we selected.

mih:
I also used the chance to (re)verify on a "dirty" windows system.

Closes #1031

### DIFF
--- a/docs/intro/installation.rst
+++ b/docs/intro/installation.rst
@@ -68,8 +68,8 @@ common scenarios, operating systems, or platforms.
    :width: 50%
    :alt: Cartoon of a person sitting on the floor in front of a laptop
 
-Windows 10
-""""""""""
+Windows 10 and 11
+"""""""""""""""""
 
 There are countless ways to install software on Windows. Here we describe *one*
 possible approach that should work on any Windows computer, like one that you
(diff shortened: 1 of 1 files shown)


docs/intro/installation.rst
```diff
@@ -96,6 +96,26 @@ Python:
     as this is required for subsequent installation steps and interactive use later on.
     Other than that, using the default installation settings is just fine.
 
+    .. windows-wit:: Very Python installation
```
adswa, line 99: Is this supposed to say "very", or something else like "verify"?

```suggestion
    .. windows-wit:: Verify Python installation
```

adswa (approved):
Thx! I just pushed a few fixes on top of your PR

Issue #1033, opened 2023-10-09
Exploration: info box on containers under windows


mih:
- install docker desktop (use WSL2-based variant)
- might need to upgrade WSL kernel version: `wsl --update`
- run docker desktop, will start the engine (this can take several minutes)
- verify function in `cmd.exe`: `docker ps`

Demo in a datalad dataset:

datalad containers-add -u dhub://debian:stable-slim debian
datalad containers-run -n debian ls

Although this works nicely, there are some issues with it: Most important https://github.com/datalad/datalad-container/issues/224 (which prevents reuse on other platforms).

adswa:
I agree its useful to have. I'll work on a draft PR

PR #1034, opened 2023-10-09
Add a FOM on docker installation on Windows

adswa:
As a fix for #1033. It would be nice to get @mih's opinion.


adswa:
Preview for easier review

image

PR #1046, opened 2023-11-01
Syntax clean

mih:
This changeset has two small changes that primarily avoid needlessly complex markup constructs.


adswa (approved):
thx!

PR #1047, opened 2023-11-01
Stop excessive colonization

mih:
I think the heading works just as fine without the prefix. Shorter is better.


adswa:
lol, "colonization"

PR #1049, opened 2023-11-01
add some words to prevent an awkward line break

adswa:
The linebreak was: .
gitmodules
image

### DIFF
--- a/docs/basics/101-139-hostingservices.rst
+++ b/docs/basics/101-139-hostingservices.rst
@@ -357,6 +357,6 @@ Once you have set up your dataset sibling(s), you can push individual datasets w
          is protecting. If the private key does not have a passphrase, simply copying
          this file grants a person access!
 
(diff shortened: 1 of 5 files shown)


docs/basics/101-139-hostingservices.rst
```diff
@@ -357,6 +357,6 @@ Once you have set up your dataset sibling(s), you can push individual datasets w
          is protecting. If the private key does not have a passphrase, simply copying
          this file grants a person access!
 
-.. [#f3]  GitHub `deprecated user-password authentication <https://developer.github.com/changes/2020-02-14-deprecating-password-auth>`_ and only supports authentication via personal access token from November 13th 2020 onwards. Supplying a password instead of a token will fail to authenticate.
+.. [#f3]  GitHub `deprecated user-password authentication <https://developer.github.com/changes/2020-02-14-deprecating-password-auth>`_ in favour of authentication via personal access token. Supplying a password instead of a token will fail to authenticate.
```
adswa, line 360:
```suggestion
.. [#f3]  GitHub `deprecated user-password authentication <https://developer.github.com/changes/2020-02-14-deprecating-password-auth>`_ in favor of authentication via personal access token. Supplying a password instead of a token will fail to authenticate.
```

PR #1050, opened 2023-11-01
Filter logs

### LINKED ISSUE #1048: Remove needless `INFO` messages and `action summary` output
These consume a lot of space, and are often not helpful given the simple usage patterns that dominate the handbook.

`autorunrecord`  has a `line_replace` feature that can be used for this, and it is already used intensely in `docs/conf.py`. For instance specific cases, it would be possible to give additional replacement specifications in the `autorunrecord` directive itself.

I would propose to remove:

- any `action summary` that does not contain `notneeded`. The exception is useful, because by default `notneeded` results are not renderer individually.
- any `INFO` log message that is not a `run` start/end output marker

  ```
  [INFO] == Command start (output follows) =====
  [INFO] == Command exit (modification check follows) ===== 
  ```

PS: replacement specifications are executed with `re.MULTILINE`

mih:
Closes #1048


adswa (approved):
much better! One of the info messages "[INFO] Attempting a clone into" still isn't caught, though...

adswa:
(merging anyway already, to see the impact on length)

PR #1052, opened 2023-11-01
Correct the description of hash-size impact on Windows paths

### LINKED ISSUE #977: [BUG] Description on windows hashing (pedantic)
### Current behavior

Minor error in hash discussion in section on Datalad on windows. It refers to the hash value length in `characters`, rather than `bits`. I don't think it is super important to fix, but it might be distracting.

https://github.com/datalad-handbook/book/blob/ab2c07c77088e369d95b654a451893ac7f5e5293/docs/intro/windows.rst?plain=1#L32-L38

And also in the footnotes

https://github.com/datalad-handbook/book/blob/ab2c07c77088e369d95b654a451893ac7f5e5293/docs/intro/windows.rst?plain=1#L168

It could simply be corrected to `bit` and, if considered useful, perhaps a mention that this 128bit is converted to a sequence of 32 hexadecimal digits in the filename (which users can then verify by thecking the git annex directory)

### Operating system

- Windows 

### DataLad information

_No response_

### Browser

- Chrome 

### Additional context

_No response_

mih:
The description is changed to state the correct sizes in characters (because those are immediately relevant, not the bitsize, even if directly convertable).

Thanks to @NemoAndrea for reporting!

Closes #977


adswa (approved):
thanks much!

PR #1062, opened 2023-11-02
Some typo fixes and punctuation tweaks

mslw:
- a few random typo or punctuation changes
- consistent spelling of macOS as macOS

### DIFF
--- a/docs/intro/installation.rst
+++ b/docs/intro/installation.rst
@@ -53,7 +53,7 @@ systems. They do not cover the various :term:`DataLad extension`s that need to b
    resulting in errors during installation and execution.
 
    But does that mean that you should uninstall Python 2?  **No**!  Keep it
-   installed, especially if you are using Linux or MacOS.  Python 2 existed for
+   installed, especially if you are using Linux or macOS.  Python 2 existed for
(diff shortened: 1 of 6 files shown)


docs/intro/installation.rst
```diff
@@ -18,7 +18,7 @@ In general, the DataLad installation requires Python 3 (see the
 why this is required), :term:`Git`, and :term:`git-annex`, and for some
 functionality `7-Zip <https://7-zip.org>`_.  The instructions below detail how
 to install the core DataLad tool and its dependencies on common operating
-systems. They do not cover the various :term:`DataLad extension`'s that need to be installed separately, if desired.
+systems. They do not cover the various :term:`DataLad extension`s that need to be installed separately, if desired.
```
adswa, line 21: I think this breaks the build, I'll toy around with it locally

adswa:
```suggestion
systems. They do not cover the various :term:`DataLad extension`\s that need to be installed separately, if desired.
```

mslw: Sorry, I didn't check the build locally. I think other places manage to work with an s but no apostrophe? In either case, the apostrophe made it to the PDF version I saw.

PR #1064, opened 2023-11-02
Add windows-wits to index

mih:
This aims to make them more accessible. The only way right now would be the list of boxes in the PDF rendering.

This is merely a first draft for relevant index entries. We need to add more to see the picture better.


adswa (approved):
this makes sense to me, thx!

PR #1065, opened 2023-11-02
Various change to indexing

mih:
This is all not yet cooked, but trying to be more consistent.


mih:
Failure is real

adswa:
I have bumped the Sphinx version in #1067, we can merge main into this branch if you want

docs/basics/101-132-advancednesting.rst
```diff
@@ -1,10 +1,12 @@
+.. index:: ! DataLad dataset; nesting
+
```
adswa, line 2: This here seems duplicated now
```suggestion
```

PR #1067, opened 2023-11-03
loosen docutils version pin to allow newer Sphinx

adswa:


adswa:
This has worked nicely, and runs through with Sphinx 7x. It has also vastly sped up dependency resolving

PR #1076, opened 2023-11-03
Start filling up the index

### LINKED ISSUE #1066: `.. index::` must go *before* section/heading
Often they are still following a headline, but this places the anchor wrong.

### LINKED ISSUE #1074: Pattern to put (FOM) boxes (and other relevant pieces) in the index
This seems to be a viable approach that doesn't require much thinking:

```rst
.. index::
   triple: howto; set a description for dataset location; with DataLad
```

This creates three index entries:

image

image

image

This covers three possible search approaches a reader may use:

- how to solve something *knowing the tool*
- *looking for any howtos*, not caring about a specific tools
- just looking for a specific use case/challenge

`howto` could also be replaced with other terms that would more or less also work and allow for a further classification of the type of book content being referenced

- `howto`: in-depth coverage of a topic
- `recipe`: compact code-centric illustration of a topic
- `example`: just some code that does the thing, not specifically described
- `note`: some kind of remark, no code, could also be `recommendation`, `warning`

`with DataLad` is a context description. Alternatives would be:

- `on Windows`
- `with Git` 
- `in the terminal`

For all three components it would be useful if the number of items in the respective super-index-item is not too large.

If the format/type of the indexed content is not particularly relevant, `pair:` can be used instead of `triple:` and the type classifier can be omitted. 

The item description should be kept broad (a reader does not necessarily know exactly what they are looking for). For example, the box "Avoid installing Python from the Windows store" should get a description as broad as `Python installation`. Combined with a context, this should provide pretty good discoverability.

Not every box needs an index entry! Often it is enough to add one to the containing section. However, large/floating boxes should get their own index markup, as they may end up many pages later.

mih:
Implements ideas from #1074 

This changeset incorporates all boxes. Other elements will come in future PR.

Various  fixes are applied to additional index items found left and right of the path.

Closes #1074
Closes #1066

### DIFF
--- a/docs/basics/101-105-install.rst
+++ b/docs/basics/101-105-install.rst
@@ -1,5 +1,6 @@
-.. index:: ! DataLad command; clone
-   pair: clone a dataset; with DataLad
+.. index::
+   pair: clone; DataLad command
+   pair: clone dataset; with DataLad
 .. _installds:
 
 Install datasets
@@ -140,7 +141,8 @@ chapters in this handbook will demonstrate how useful this information can be.
 
 Here is the repository structure:
 
-.. index:: terminal command; tree
+.. index::
+   pair: tree; terminal command
    pair: display directory tree; on Windows
 .. windows-wit:: use tree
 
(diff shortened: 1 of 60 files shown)


docs/basics/101-105-install.rst
```diff
@@ -305,6 +304,8 @@ that the download of one file was ``notneeded`` and the retrieval of the other f
 
    :dlcmd:`get` uses :gitannexcmd:`get` underneath the hood.
 
+.. index::
+   pair: drop files; with DataLaD
```
adswa, line 308:
```suggestion
   pair: drop files; with DataLad
```

mih: thx!

PR #1100, opened 2023-11-06
Add commas to 'for example', where applicable

### LINKED ISSUE #1084: Set commas for `for example` properly
- comma before when a parenthetical statement starts
- comma after, when a actual example follows

adswa:
Fixes #1084. I did not add commas to absolutely every occurrence, because (the rule can be loosened when it would affect readability](https://english.stackexchange.com/questions/132359/any-exception-with-commas-before-and-after-for-example). In some instances, I also simplified sentences or replaced 'for example' with 'such as'. But its not a trivial task for a non-native speaker, I feel like I lack the necessary "Sprachgefühl" to be sure about commas.


mih (approved):
LGTM! Thx!

PR #1112, opened 2023-11-06
Move box content into dedicated file

mih:
This is a minimal test change to see how it looks.

The main goal is to trim the diff to the print version again, where content was moved and taken out of containers. This makes merged conflict-ladden and painful. With this approach, any content change would happen in the dedicated file, which can be used for boxes and sections alike. Associated index items must stay outside, because they are declared on the container.


adswa (approved):
I think that's an elegant solution. Thx! I went through the PR and checked that linked files contained the right content

PR #1142, opened 2023-11-09
turn 'hidden section' references into find-out-more refs

adswa:
to fix https://github.com/psychoinformatics-de/book-datalad-intro/issues/103


adswa:
ah, the failure is legit... I believe its because I added ``:name:`` options to findoutmores, but did not make them ``:float:``s... What would be your advice, @mih, make them floats or get creative with the description?

Issue #1127, opened 2023-11-09
Collaboration chapter `whereis` demo broken


mih:
It discusses a different between files (two vs three sources), but the logs are identical for both.

adswa:
That is true. I am inclined to resolve this section and demo "whereis" maybe on a retrieved and unretrieved file

adswa:
On closer look, it doesn't really add anything new to the previous chapter's ``whereis`` exploration on the different books. My suggestion would be to move the ``whereis`` part of ``101-116-sharelocal.rst`` into ``101-117-sharelocal2.rst``. It helps to reduce the length of the first "Share local" chapter and keeps the "Where's Waldo" chapter, and does a good job of demonstrating ``whereis``

mih:
done, thx

PR #1144, opened 2023-11-09
Move git-annex whereis content, remove dysfunctional demo

adswa:
This moves a chunk of text about ``git annex whereis`` from the "Looking without touching" chapter into "Where's Waldo", and replaces a demo that became dysfunctional over the years (#1127). This PR needs an appveyor run because new example code snippets need to be build.


mih:
Merging blind. Hard to confirm without the full context. Needs a proper read.

PR #1177, opened 2023-11-13
For these snippets we need to see a UUID

mih:
otherwise the text does not really make sense.


adswa (approved):
fine with me, but I don't think it will survive a rerun on appveyor, right?

PR #1183, opened 2023-11-14
Adjusted mode

### LINKED ISSUE #1136: Explicit box on "adjusted mode"
The box already exists, but is called "What happens on windows?", and does not mention the term "adjusted mode", but still describes most of what it is.

We could do one or more of the following:

- add glossary term `adjusted mode` -- the name is already used and a topic item also is in use already
- add a box on adjusted mode, and that the files in the worktree are pointer files instead of symlinks, and that that is the reason we access works and does not fail in the same whay as with symlinks (albeit the content of such files being unexpected), etc.
- rename/rephrase the existing box to be explicitly about `adjusted mode`
- mention that `adjusted mode` is not per se a Windows thing, but is determined by the file system (think FAT32 usb stick on a Mac)

mih:
Explicit mention in box, new glossary term (and also one for 'corresponding branch'), plus minor updates for the 'adjusted branch' term.

Add a hint that this is not at all windows-only.

Closes #1136

### DIFF
--- a/docs/basics/topic/adjustedmode-nosymlinks.rst
+++ b/docs/basics/topic/adjustedmode-nosymlinks.rst
@@ -1,6 +1,6 @@
 Windows has insufficient support for :term:`symlink`\s and revoking write :term:`permissions` on files.
 Therefore, :term:`git-annex` classifies it as a :term:`crippled file system` and has to stray from its default behavior: it enters :term:`adjusted mode`.
-While git-annex on Unix-based file operating systems stores data in the annex and creates a symlink in the data's original place, on Windows it moves data into the :term:`annex` and creates a *copy* of the data in its original place. This behavior is not specific to Windows, but is done for any impaired file system, such a dataset on a USB-stick plug into a Mac.
+While git-annex on Unix-based file operating systems stores data in the annex and creates a symlink in the data's original place, on Windows it moves data into the :term:`annex` and creates a *copy* of the data in its original place. This behavior is not specific to Windows, but is done for any impaired file system, such as a dataset on a USB-stick plugged into a Mac.
 
 **Why is that?**
 Data *needs* to be in the annex for version control and transport logistics -- the annex is able to store all previous versions of the data, and manage the transport to other storage locations if you want to publish your dataset.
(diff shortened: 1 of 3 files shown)


docs/basics/topic/adjustedmode-nosymlinks.rst
```diff
@@ -1,6 +1,6 @@
 Windows has insufficient support for :term:`symlink`\s and revoking write :term:`permissions` on files.
-Therefore, :term:`git-annex` classifies it as a :term:`crippled file system` and has to stray from its default behavior.
-While git-annex on Unix-based file operating systems stores data in the annex and creates a symlink in the data's original place, on Windows it moves data into the :term:`annex` and creates a *copy* of the data in its original place.
+Therefore, :term:`git-annex` classifies it as a :term:`crippled file system` and has to stray from its default behavior: it enters :term:`adjusted mode`.
+While git-annex on Unix-based file operating systems stores data in the annex and creates a symlink in the data's original place, on Windows it moves data into the :term:`annex` and creates a *copy* of the data in its original place. This behavior is not specific to Windows, but is done for any impaired file system, such a dataset on a USB-stick plug into a Mac.
```
adswa, line 3:
```suggestion
While git-annex on Unix-based file operating systems stores data in the annex and creates a symlink in the data's original place, on Windows it moves data into the :term:`annex` and creates a *copy* of the data in its original place. This behavior is not specific to Windows, but is done for any impaired file system, such as a dataset on a USB-stick plugged into a Mac.
```

docs/glossary.rst
```diff
@@ -127,6 +146,15 @@ Glossary
       A text file that lists all required components of the computational environment that a :term:`software container` should contain.
       It is made by a human user.
 
+   corresponding branch
+      .. index::
+         pair: corresponding branch; in adjusted mode
+
+      A :term:`branch` underlying a particular :term:`adjusted branch`.
+      Changes committed to an adjusted branch are progated to its corresponding
```
adswa, line 154:
```suggestion
      Changes committed to an adjusted branch are propagated to its corresponding
```

adswa (approved):

PR #1185, opened 2023-11-14
On annex-keys

### LINKED ISSUE #1132: git-annex chapter dances around the concept of a annex-key
It is explained in a box, but more frequently the key is referred to as a "checksum" -- which is mostly, but not entirely, correct.

I am pondering to:

- turn any mention of "key" into "annex-key" to make more obvious that this is a special term
- add a glossary item for "annex-key"
- replace mentions of "checksum" with "annex-key", whenever it is really about the identifier aspect and less about the hash
- introduce a "annex-backend" term too -- presently `MD5E` is labeled a "checksum"

mih:
Explicit term. More acurrate description, less focused on checksums only (there are checksum-less backends).

Closes #1132

### DIFF
--- a/docs/basics/101-115-symlinks.rst
+++ b/docs/basics/101-115-symlinks.rst
@@ -194,7 +194,7 @@ will take a closer look at that.
    This checksum *uniquely* identifies a file's content.
    A hash function will generate the same character sequence for the same file content, and once file content changes, the generated checksum changes, too.
    Basing the file name on its contents thus becomes a way of ensuring data integrity:
-   File content cannot be changed without git-annex noticing, because file's checksum, and thus its key in its symlink, will change.
+   File content cannot be changed without git-annex noticing, because the file's checksum, and thus its key in its symlink, will change.
    Furthermore, if two files have identical checksums, the content in these files is identical.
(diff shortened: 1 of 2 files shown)


docs/basics/101-115-symlinks.rst
```diff
@@ -175,30 +175,30 @@ will take a closer look at that.
...
+   computed from some input, for example the content of a PDF file,
+   by a *hash* function.
 
-   Importantly, a hash function will generate the same character sequence for the same file content, and once file content changes, the generated hash changes, too.
+   This checksum *uniquely* identifies a file's content.
+   A hash function will generate the same character sequence for the same file content, and once file content changes, the generated checksum changes, too.
    Basing the file name on its contents thus becomes a way of ensuring data integrity:
-   File content cannot be changed without git-annex noticing, because file's hash, and thus its key in its symlink, will change.
-   Furthermore, if two files have identical hashes, the content in these files is identical.
+   File content cannot be changed without git-annex noticing, because file's checksum, and thus its key in its symlink, will change.
```
adswa, line 197:
```suggestion
   File content cannot be changed without git-annex noticing, because the file's checksum, and thus its key in its symlink, will change.
```

PR #1189, opened 2023-11-14
mention datalad-configuration command

adswa:


docs/basics/101-122-config.rst
```diff
@@ -295,6 +295,21 @@ remaining sections in that file, and the :ref:`that dissects this config file fu
...
 
+.. index::
+   pair: configuration; DataLad command
+   pair: setting configurations; with DataLad
+
+The ``datalad configuration`` command
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Although this section put a focus on the ``git config`` command, it is important to mention that there also is a ``datalad configuration`` command.
+It is not identical to ``git config``, but while it lacks some feature of ``git config``, such as the ability to set system-wide configuration, it has additional features:
```
mih, line 306:
```suggestion
It is not identical to ``git config``, but while it lacks some feature of ``git config``, such as the ability to set system-wide configuration, it has additional features.
```

I am removing such colons. To me they only serve a purpose within a sentence, but less so across.

docs/basics/101-122-config.rst
```diff
@@ -295,6 +295,21 @@ remaining sections in that file, and the :ref:`that dissects this config file fu
...
+Beyond the ``local`` and ``global`` scopes, it also supports :term:`branch` specific configurations in the ``.datalad/config`` file (further discussed in the next section), setting configurations recursively through dataset hierarchies, and multi-configuration queries (such as ``datalad configuration get user.name user.email``).
+By default, ``datalad configuration`` will ``dump`` (list) the effective configuration including relevant ``DATALAD_*`` :term:`environment variable`\s.
```
mih, line 308:
```suggestion
By default, ``datalad configuration`` will ``dump`` (list) the effective configuration including relevant ``DATALAD_*`` :term:`environment variable`\s, and also annotate the purpose of many common configuration items.
```

docs/basics/101-122-config.rst
```diff
@@ -295,6 +295,21 @@ remaining sections in that file, and the :ref:`that dissects this config file fu
...
+The subcommands ``datalad configuration get`` or ``datalad configuration set`` perform queries or set configurations.
+You can find out more information on this command in the `command documentation <https://docs.datalad.org/en/stable/generated/man/datalad-configuration.html>`_.
```
mih, line 310: I believe this linkage is what the `dlcmd` directive would do (for HTML). I am not sure if we tend to link standard documentation for commands. If not, we should also not start it here.

mih (commented):
Made some style suggestions. thx!

docs/basics/101-122-config.rst
adswa: done

PR #1204, opened 2023-12-10
No more beta

mih:
I think with a physical book out, it is time to remove the statement from the frontpage. The note is kept, though, because the spirit should stay the same.


adswa (approved):
alrighty!

PR #1219, opened 2024-04-22
simple changes

CPernet:
add some windows subsystem info  
fix some comma issue for DataCat example


jsheunis (changes requested):
Thanks for the PR @CPernet. Two points:
- The addition of the commas in the catalog file are fine, those can be merged.
- The additional note about WSL2 for windows installation is actually already covered in a note further down on that same page:

```
.. windows-wit:: Install DataLad using the Windows Subsystem 2 for Linux
   :name: ww-wsl2

   With the Windows Subsystem for Linux, you will be able to use a Unix system
   despite being on Windows.  You need to have a recent build of Windows in
   order to get WSL2 -- we do not recommend WSL1.

   You can find out how to install the Windows Subsystem for Linux at
   `docs.microsoft.com <https://learn.microsoft.com/en-us/windows/wsl/install>`_.
   Afterwards, proceed with your installation as described in the installation instructions
   for Linux.
```

So I would say the change to `installation.rst` is not necessary. Could you remove that commit from the PR?

CPernet:
pushing a clean one instead

PR #1231, opened 2024-07-04
Fixes to the S3 walk-through

### LINKED ISSUE #1224: Issues with walkthrough for S3 special remote
Hi, thank you for Datalad and the handbook.

I am working through the[ walkthrough on setting up S3 as a special remote](https://handbook.datalad.org/en/latest/basics/101-139-s3.html), using my own small dataset.

When I run the `git annex initremote` command as shown there, I get an error.

```
$ git annex initremote public-s3 type=S3 encryption=none \
> bucket=$BUCKET public=yes datacenter=EU autoenable=true 
initremote public-s3 (checking bucket...) (creating bucket in EU...) 
git-annex: S3Error {s3StatusCode = Status {statusCode = 400, statusMessage = "Bad Request"}, s3ErrorCode = "InvalidBucketAclWithObjectOwnership", s3ErrorMessage = "Bucket cannot have ACLs set with ObjectOwnership's BucketOwnerEnforced setting", s3ErrorResource = Nothing, s3ErrorHostId = Just "[possibly sensitive info redacted]", s3ErrorAccessKeyId = Nothing, s3ErrorStringToSign = Nothing, s3ErrorBucket = Nothing, s3ErrorEndpointRaw = Nothing, s3ErrorEndpoint = Nothing}
failed
initremote: 1 failed
```

I think this is the same issue as described here on the git annex wiki that is linked to from that chapter of the handbook:
https://git-annex.branchable.com/special_remotes/S3/#comment-fcfba0021592de4c1425d3bf3c9563d3

> ACL deprecation vs public=yes
> 
> Amazon has [deprecated ACLs](https://docs.aws.amazon.com/AmazonS3/latest/userguide/about-object-ownership.html)
> 
>     A majority of modern use cases in Amazon S3 no longer require the use of ACLs, and we recommend that you disable ACLs except in unusual circumstances where you need to control access for each object individually. With Object Ownership, you can disable ACLs and rely on policies for access control. When you disable ACLs, you can easily maintain a bucket with objects uploaded by different AWS accounts. You, as the bucket owner, own all the objects in the bucket and can manage access to them using policies.
> 
> They are encouraging everyone to [migrate to bucket policies](https://docs.aws.amazon.com/AmazonS3/latest/userguide/object-ownership-migrating-acls-prerequisites.html) instead.

I haven't wrapped my head around the AWS policy-ese they outline in the suggested fix but I thought I should go ahead and report the issue here

adswa:
This incorporates the corrected commands and information from #1224.
Fixes #1224.

### DIFF
--- a/docs/basics/101-139-s3.rst
+++ b/docs/basics/101-139-s3.rst
@@ -222,6 +222,28 @@ To make them public, switch to the "Permissions" tab in your buckets S3 console 
 
    Bucket settings allow making the bucket public
 
+Alternatively, create a bucket policy as shown below, inserting your own bucket name into the two placeholders::
+
+    {
+    "Version": "2012-10-17",
+    "Statement": [
+        {
+            "Effect": "Allow",
+            "Principal": "*",
+            "Action": "s3:GetObject",
+            "Resource": [
+                "arn:aws:s3:::YOUR-BUCKET-NAME-HERE",
+                "arn:aws:s3:::YOUR-BUCKET-NAME-HERE/*"
+                ]
+            }
+        ]
+    }
+
+.. figure:: ../artwork/src/aws_s3_bucket_policy.png
+
+   Bucket policy to allow objects in the bucket to be retrieved by anyone.
+
+
 .. find-out-more:: Info on public buckets created prior to April 2023
 
     Amazon S3 buckets created before April 2023 supported using ACLs for public read access to files.
--- a/docs/basics/101-139-s3.rst
+++ b/docs/basics/101-139-s3.rst
@@ -3,6 +3,12 @@
 Walk-through: Amazon S3 as a special remote
 -------------------------------------------
 
+.. importantnote:: This walk-through requires git-annex >= 10.20230802
+
+   Prior versions of git-annex do not support public access via the ``publicurl`` parameter with S3 buckets created after April 2023.
+   Find out more about this in `this discussion <https://git-annex.branchable.com/bugs/S3_ACL_deprecation/>`_.
+
+
 `Amazon S3 <https://aws.amazon.com/s3>`_ (or Amazon Simple Storage Service) is a
 popular service by `Amazon Web Services <https://aws.amazon.com>`_ (AWS) that
 provides object storage through a web service interface. An S3 bucket can be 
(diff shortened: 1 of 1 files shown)


adswa:
hey @all-contributors please add @NickleDave for bug, content

docs/basics/101-139-s3.rst
```diff
@@ -209,6 +209,22 @@ to "Buckets" to see your newly created bucket. It should only have a single
...
+By default, this bucket and its contents are not publicly accessible.
+To make them public, switch to the "Permissions" tab in your buckets S3 console overview, and turn the option "Block all public access" off.
+
+.. figure:: ../artwork/src/aws_s3_bucket_permissions.png
+
+   Bucket settings allow making the bucket public
+
+.. find-out-more:: Info on public buckets created prior to April 2023
+
+    Amazon S3 buckets created after April 2023 had support for using ACLs for public read access to files.
```
christian-monch, line 221:
```suggestion
    Amazon S3 buckets created before April 2023 supported using ACLs for public read access to files.
```

mslw (changes requested):
I believe we should also mention the `publicurl` git-annex parameter. Otherwise, I think git-annex would use s3 protocol, and I am not sure if it does it anonymously. See its description in [git-annex S3 manpage](https://git-annex.branchable.com/special_remotes/S3/).

There is, however, a caveat that the git-annex available currently in debian-stable does not recognize the `publicurl` parameter, and blindly follows `public` :see_no_evil: I think this is worth another find-out-more. See also [this comment](https://github.com/datalad-handbook/book/issues/1224#issuecomment-2155481884).

adswa:
> There is, however, a caveat that the git-annex available currently in debian-stable does not recognize the publicurl parameter, and blindly follows public 🙈 I think this is worth another find-out-more. See also https://github.com/datalad-handbook/book/issues/1224#issuecomment-2155481884.

I have added a version note to the top of the page, and I also linked the discussion in it. 

> I believe we should also mention the publicurl git-annex parameter. Otherwise, I think git-annex would use s3 protocol, and I am not sure if it does it anonymously. See its description in [git-annex S3 manpage](https://git-annex.branchable.com/special_remotes/S3/).

Do you mean beyond the description prior to `enableremote`?

I realize I haven't mentioned the bucket policies - I will read up on what they are and add a findoutmore on them.

mslw:
Ah, sorry, `publicurl` has already been there, I focused on the diff, and the part which removed `public` -- but at that stage in the chapter we are assuming that everything is private.

So it isn't exactly as bad as what I said (that older annex does not recognize publicurl at all -- it does not recognize it when `public=no`). You can check again whether the important note at the top is not too strict (because of me), but otherwise I retract my request for changes :-)

mslw (approved):

PR #1234, opened 2024-07-17
TMP CI RTD 

adswa:
add an arbitrary change just to see if readthedocs breaks with addons enabled but not conf.py change


adswa:
served its purpose: even without the new recommended configuration change (#1233) and enabled addons in the RTD projects, builds look fine.

PR #1235, opened 2024-09-06
Fixes for minor things I noticed while reading the handbook

matrss:
Just some minor fixes for things I noticed reading through the first and second chapter of "Basics".


adswa (approved):
thank you very much!

PR #1239, opened 2024-09-10
add example terraform config for an s3 bucket

mikapfl:
Adding public access to an S3 bucket for a publicly available datalad repository via opentofu is quite a few lines of configuration because you essentially have to configure it twice.

I put this into the "Advanced example" category because it is only useful for people using terraform / opentofu.


adswa (approved):
Thank you very much! I'm unfamiliar with terraform or opentofu, but that shouldn't stop this snippet from going in

adswa:
hey @all-contributors please add @mikapfl for content

PR #1245, opened 2024-12-27
replace word 'crop' with 'resize'

JoKeyser:
Hi, I think the word "crop" in this sentence may be slightly misleading and should be replaced with "resize".
To me the image operations of cropping and resizing are very different.
Since the text and examples beyond that sentence all talk about resizing, it should also be called resizing here as well.

### DIFF
--- a/docs/basics/101-110-run2.rst
+++ b/docs/basics/101-110-run2.rst
@@ -324,7 +324,7 @@ the ``-o``/``--output`` option.
    pair: declare command output; with DataLad run
 
 In order to execute :dlcmd:`run` with both the ``-i``/``--input`` and ``-o``/``--output``
-flag and see their magic, let's crop the second logo, ``logo_interval.jpg``:
+flag and see their magic, let's resize the second logo, ``logo_interval.jpg``:
 
 .. index::
    pair: files are unlocked by default; on Windows


malikwirin:
Doesn't cropping just mean to resize it for it to become smaller? If so it would be a good word to avoid repetition

JoKeyser:
> Doesn't cropping just mean to resize it for it to become smaller? If so it would be a good word to avoid repetition

Hi, thanks for checking in.
I think the word cropping refers to "the removal of unwanted outer areas from an image", like [written in wikipedia](https://en.wikipedia.org/wiki/Cropping_(image)). That will also reduce the size of the image, and thus is related, but it's an entirely different operation.
I think it's both confusing to users who know the difference and those who don't; it may only not be confusing to those who think the two concepts are the same :).

adswa:
Thanks both for the PR and short discussion here! I am not particularly attached to the word, I would be fine changing it. 


@all-contributors please add @malikwirin for review 

@all-contributors please add @JoKeyser for bug

@JoKeyser feel free to add yourself to the .zenodo.json file of this repository, in the second to last position (before Michael Hanke), or leave your name, orcid and affiliation in a comment and I'll add you. :)

adswa (approved):

JoKeyser:
@adswa , thank you very much, I've opened a PR https://github.com/datalad-handbook/book/pull/1250.

PR #1250, opened 2025-01-02
add myself to zenodo.json

JoKeyser:
Hi, as @adswa [suggested here](https://github.com/datalad-handbook/book/pull/1245#issuecomment-2566531250), I'd like to be added to the list of co-creators.
Thank you for keeping track so diligently.


adswa:
Thanks!

PR #1253, opened 2025-03-09
Set matplotlib backend as non interactive

malikwirin:
Code is assuming to have a X display server available, which caused an [issue](https://github.com/datalad/datalad-container/issues/272) for me.
This PR will set the matplotlib backend as non interactive which fixed the issue in my testing.


adswa (approved):
thanks for the PR and the in-depth analysis of the issue! I think this is a sensible change.

adswa:
@all-contributors please add @malikwirin for bug, code

PR #1258, opened 2025-06-08
TMP: remove git-annex installation

adswa:
For the CI runs, we should install `git-annex` and `datalad` as individual tools (`uv tool install`), and avoid mixing the deployment with the handbooks internal dependencies. This should give us a more realistic deployment (compared to a user system, or one we'd use for a workshop).


adswa:
The Appveyor failures relate to the examples from the challenges. At the moment it is:

```
    download(error): chinstrap_02.jpg [download failure] [403 Client Error: Forbidden for url: https://unsplash.com/photos/8PxCm4HsPX8/download?force=true]

```

Which I don't see locally. My suspicion is rate limiting by unsplash.

PR #1262, opened 2025-06-11
Tune test setup

mih:
This aims to remove some unnecessary complexity from the setup and make it more widely applicable.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -26,6 +26,7 @@ shallow_clone: false
 environment:
   DTS: datalad_next
   APPVEYOR_BUILD_WORKER_IMAGE: Ubuntu2004
+  AUTORUNRECORD_BASEDIR: /home/appveyor/workdir
   INSTALL_SYSPKGS: graphicsmagick-imagemagick-compat moreutils jq uidmap fuse2fs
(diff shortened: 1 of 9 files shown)


.appveyor.yml
```diff
@@ -25,8 +25,8 @@ shallow_clone: false
 
 environment:
   DTS: datalad_next
-  # 20.04 would require a 2.31 GLIBC, but git-annex is only built for 2.34
-  APPVEYOR_BUILD_WORKER_IMAGE: Ubuntu2204
+  APPVEYOR_BUILD_WORKER_IMAGE: Ubuntu2004
```
adswa, line 28: I think there was no good reason other than "it was like this in the beginning and we haven't updated it since" for this older Ubuntu. Thanks for reinstating it, but I think we can keep it more modern and stay with 22.04

adswa (commented):
Thank you! I'm letting this run through locally still, but it looks good. I'd only propose to drop 04ac1f17e9c62c2723d220939cf92a60ea16f547 (see my comment)

adswa:
First observation: A users gitconfig file alters the output. E.g. I still have my git configured to use ``master`` (residue from Gin use) and that makes the build fail quite early (its is expecting ``main`` in a sibling)

PR #1285, opened 2026-07-15
New pypi install instructions for datalad-slurm

knuedd:
Now that datalad-slurm can be installed with pypi, update it in the handbook


adswa (approved):
🎉  Thank you! :)

